PR #41, opened 2022-05-04
Update readme 

mchaynes:
Update `README.md`

Will update links to point to registry and include java installation instructions when those are done.

Filed https://github.com/pulumi/pulumi-pulumiservice/issues/40 to track adding CONTRIBUTING.md

PR #43, opened 2022-05-04
Update python/README.md 

mchaynes:
Merged in the readme from #41 without realizing that we had failures in the build. Looks like we committed the sdk/python/README.md to git and so when the codegen ran we updated it to the new readme. So instead of committing two copies, lets just remove it and add it to the .gitignore

Failing run from previous PR: https://github.com/pulumi/pulumi-pulumiservice/runs/6284621845?check_suite_focus=true


lukehoban:
FWIW - for better or worse, I believe this is intentional, and should in generally not be gitignored. Instead, you have to make sure to rebuild SDKs after readme changes. Cc @stack72.

mchaynes:
> FWIW - for better or worse, I believe this is intentional, and should in generally not be gitignored. Instead, you have to make sure to rebuild SDKs after readme changes. Cc @stack72.

Interesting. Ok! I'll regen and add the readme back then

EvanBoyle (approved):

Issue #58, opened 2022-05-16
YAML Programs fail with error


mchaynes:
When attempting to run a `pulumi up` on a YAML program, we get an error.

```
Diagnostics:
  pulumi:pulumi:Stack (github-secrets-dev):
    error: Error: error resolving type of resource webhook: internal error loading package "pulumiservice": readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...

      on Pulumi.yaml line 13:
      13:     type: pulumiservice:Webhook

    error resolving type of resource webhook: internal error loading package "pulumiservice": readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
    error: an unhandled error occurred: failed to evaluate template

    Error: error resolving type of resource webhook: internal error loading package "pulumiservice": readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
      on Pulumi.yaml line 13:
      13:     type: pulumiservice:Webhook
    error resolving type of resource webhook: internal error loading package "pulumiservice": readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...

```

`./Pulumi.yaml`
```yaml
name: example-service-yaml
runtime: yaml
description: pulumi service yaml example

resources:
  webhook:
    type: pulumiservice:Webhook
    properties:
      active: true
      displayName: webhook
      payloadUrl: https://example.com/webhook
outputs: {}
```

PR #62, opened 2022-05-20
add YAML support

mchaynes:
## Details
Adds YAML support for the Pulumi Service Provider. Originally was under the impression that YAML would work out of the box, however YAML requires the `GetSchema` GRPC call to return the actual schema, since otherwise YAML has no way of determining the schema. The other languages worked just fine because they have SDKs which already know the schema.

Fixes: https://github.com/pulumi/pulumi-pulumiservice/issues/58

## Testing
1. `$ cd examples/yaml-teams`
2. `$ pulumi up`

Additionally, added test runner for yaml-teams examples and github actions config as part of this PR.

### DIFF
--- a/provider/cmd/pulumi-resource-pulumiservice/main.go
+++ b/provider/cmd/pulumi-resource-pulumiservice/main.go
@@ -15,12 +15,19 @@
 package main
 
 import (
+	_ "embed"
+
 	"github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider"
 	"github.com/pulumi/pulumi-pulumiservice/provider/pkg/version"
 )
 
 var providerName = "pulumiservice"
 
+// embed schema.json directly into resource binary so that we can properly serve the schema
+// directly from the resource provider
+//go:embed schema.json
+var schema string
+
 func main() {
-	provider.Serve(providerName, version.Version)
+	provider.Serve(providerName, version.Version, schema)
 }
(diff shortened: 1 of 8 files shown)


EvanBoyle (approved):
Overall looks good!

I noticed that the Java test suite is also running the YAML tests for some reason. Once CI is green this should be good to go though!

provider/cmd/pulumi-resource-pulumiservice/main.go
```diff
@@ -15,12 +15,19 @@
...
 	"github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider"
 	"github.com/pulumi/pulumi-pulumiservice/provider/pkg/version"
 )
 
 var providerName = "pulumiservice"
 
+// embed schema.json directly into resource binary so that we can properly serve the schema
+// directly from the resource provider
+//go:embed schema.json
+var schema string
```
mchaynes, line 29: Note: `schema.json` is under the exact same path as `main.go`

```
provider/cmd/pulumi-resource-pulumiservice/main.go
provider/cmd/pulumi-resource-pulumiservice/schema.json
```

mchaynes:
> I noticed that the Java test suite is also running the YAML tests for some reason. Once CI is green this should be good to go though!

Good catch. looks like i messed up the build tag comment. did `// go:build yaml` instead of `//go:build yaml` 😵‍💫

Issue #51, opened 2022-05-10
Support team-stack permissions


mchaynes:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

@MitchellGerdisch commented on the initial request to create the pulumi service provider to add team-stack permissions:
> The provider should support team-stack permissions assignment management

https://github.com/pulumi/service-requests/issues/18#issuecomment-1017599144

sinePaul:
Hello, 
I will be interested to implement this feature :)

PR #72, opened 2022-07-08
WIP Stack permissions for teams

jeid64:
Howdy, it's been probably a year since I last touched golang/a go Pulumi provider

We need https://github.com/pulumi/pulumi-pulumiservice/issues/51 because managing perms with the Pulumi service is not great, so I've started working on adding support for it. I don't handle deleting permissions, validating that the permission selected is valid (103, 102, 101), comparing permissions to see if its changed between old/new, and I don't think the schema looks great for it. 

Posting this to start getting comments on it and see if Pulumi would like to take it over.

### DIFF
--- a/provider/cmd/pulumi-resource-pulumiservice/schema.json
+++ b/provider/cmd/pulumi-resource-pulumiservice/schema.json
@@ -91,6 +91,24 @@
         "organizationName": {
           "description": "The organization's name.",
           "type": "string"
+        },
+        "stackPermissions": {
+          "description": "A list of stacks to give the team access to",
+          "type": "array",
+          "items": {
+            "type":"object",
+            "properties": {
+              "projectName": {
+                "type": "string"
+              },
+              "stackName": {
+                "type": "string"
+              },
+              "permission": {
+                "type": "number"
+              }
+            }
+          }
         }
       },
       "inputProperties": {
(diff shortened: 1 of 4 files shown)


mchaynes:
Hi! Thank you for submitting this! We are definitely interested in completing this feature as soon as we can schedule it. I'll check in with the team and see when we can get this work prioritized.

mchaynes:
Hi, just wanted to let you know that I've begun working on this. I hit a snag in pulumi's codegen, so this will be blocked until https://github.com/pulumi/pulumi/issues/9813 is resolved.

mchaynes:
Hello! Thanks again for the PR. I'm closing out this pull request in favor of https://github.com/pulumi/pulumi-pulumiservice/pull/81, which just merged (You're still a coauthor on the PR 😄). The publish SDK job is not quite done running, but it will be released in v0.1.6 shortly.

You can see an example of usage of this resource [here](https://github.com/pulumi/pulumi-pulumiservice/blob/main/examples/yaml-team-stack-permissions/Pulumi.yaml). The implementation I went with is slightly different from yours, mainly that it creates a separate resource, instead of adding an array field to the team resource. This allows us to add stack permissions to a team without needing to directly access that team resource, which means you can manage team permissions for a stack inside of the stack itself.

Issue #75, opened 2022-07-22
StackTag values cannot be changed, appearing to corrupt stack state


christophermaier:
### What happened?

I'm setting up a handful of tags on our stacks, but encountered issues when trying to change the _value_ of a tag.

The first `pulumi up` run after changing the value looks normal: the change is detected, the diff is correct, and `pulumi up` succeeds. However, the actual value of the tag on the Pulumi SaaS has not been changed. Furthermore, subsequent `pulumi up` runs show that the tag still needs to be updated, but the diff is empty. This persists indefinitely.

Attempting to just delete the StackTag resource entirely and starting over results in persistent errors:
```
error: field type "string" does not match property "invalid"
```
The only way to recover from this appears to be manually removing the resource from the stack with `pulumi state delete`, and manually fixing the tag via the SaaS UI.

### Steps to reproduce

I can reliably reproduce this with the following YAML project (update the stack parameters as appropriate for the stack you wish to tag):
```yaml
name: pulumi_service_provider_bug
runtime: yaml
description: A minimal Pulumi YAML program
resources:
  testing-tag:
    type: pulumiservice:StackTag
    properties:
      organization: <YOUR_ORGANIZATION>
      project: <YOUR_PROJECT>
      stack: <YOUR_STACK>
      name: testing-tag
      value: "foo"
```
Put that into an empty directory and run `pulumi up`; the tag will be set up as expected.

Now, change the `value` property to, say, `"bar"`. Run `pulumi up` again. It succeeds, but the value isn't actually changed.

Run `pulumi up` again. Pulumi indicates there's an update to be made, but the diff shows nothing.

Comment out the entire `resources` object and run `pulumi up`; this _should_ remove the tag entirely. However, it fails with the above error message.

Run `pulumi up` repeatedly and observe the same failure.

### Expected Behavior

I would expect to be able to change the value of a StackTag without issue.

### Actual Behavior

The value is never changed, and subsequent attempts to delete the tag appear to corrupt the entire stack state, preventing further updates from succeeding.

### Versions used

```
CLI          
Version      3.36.0
Go Version   go1.17.12
Go Compiler  gc

Plugins
NAME           VERSION
pulumiservice  unknown
yaml           unknown

Host     
OS       debian
Version  11.4
Arch     x86_64
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

[...]

PR #76, opened 2022-07-25
Implement Update and Fix Missing Property Bug

mchaynes:
### Description of Change

This PR fixes https://github.com/pulumi/pulumi-pulumiservice/issues/75.

There were two issues that needed to be fixed to get the issue resolved.

1. Implementing update in stack tags resource
2. Fix bug where deleting resources causes the following error: `error: field type "string" does not match property "invalid"`. This was caused by our `serde` package assuming that the PropertyMap included all entries.

### DIFF
--- a/examples/examples_yaml_test.go
+++ b/examples/examples_yaml_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/pulumi/pulumi/pkg/v3/engine"
 	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
+	"gopkg.in/yaml.v2"
 )
 
 func TestYamlTeamsExample(t *testing.T) {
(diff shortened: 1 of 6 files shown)


examples/examples_yaml_test.go
```diff
@@ -23,3 +24,40 @@ func TestYamlTeamsExample(t *testing.T) {
...
+		Dir:         path.Join(cwd, ".", "yaml-stack-tags"),
+		StackName:   "test-stack",
+		PrepareProject: func(_ *engine.Projinfo) error {
+			return nil
+		},
+		EditDirs: []integration.EditDir{
+			{
+				Dir: tmpdir,
+			},
+		},
```
blampe, line 61: Just to make sure I understand - the `Dir` sets your initial resources to the stuff you defined in `examples/yaml-stack-tags/Pulumi.yaml`, then the `EditDirs` is basically simulating an updating using the resource-free file you just wrote?

blampe (approved):

examples/examples_yaml_test.go
mchaynes: That is correct, yes!

Issue #73, opened 2022-07-08
Specifying a non-existent member during Team resource creation causes the Team to be marked failed but doesn't delete the team 


jeid64:
### What happened?

I created a new Pulumi stack and wanted to make a new Team. I passed a non-existent member string to Team which caused a 
"failed to update team membership: 404 API error: Not Found: User 'julien.eid@dutchie.com' not found" error to occur. 

I then fixed the member string to be an existing member and ran pulumi up again, which failed because the team now existed. 
```
 +   └─ pulumiservice:index:Team  team                                          **creating failed**     1 error

Diagnostics:
  pulumiservice:index:Team (team):
    error: error creating team 'jeidtestteamprovider': failed to create team: 409 API error: Conflict: Team with name "jeidtestteamprovider" already exists
```
    
I had to go to the Pulumi UI and delete the team by hand, which then allowed my "Team" resource to be created. This is because the provider calls CreateTeam first here https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/team.go#L199 and then adds members to the team here https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/team.go#L205 

### Steps to reproduce

```
team = Team(
    "team",
    name="teamnamehere",
    organization_name="service-provider-test-org",
    display_name="PulumiUP Team",
    team_type="pulumi",
    members=["obviouslynonexistingmemberhere"],
)
```

### Expected Behavior

If the addmembers call fails, delete the team so that the next Team creation works.

### Actual Behavior

Left me with a team that exists but I'm unable to actually use the Pulumi Team resource because it now has a 409

### Additional context

The API around adding members to teams is a bit weird. I think the Team resource should not accept "members" at all, and there should be a separate resource that takes a Team as an Input and a member name and then calls the API to do it. 

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mchaynes:
Hi, thank you for submitting this bug report! I'll speak with the team about getting this work prioritized.

mchaynes:
> The API around adding members to teams is a bit weird. I think the Team resource should not accept "members" at all, and there should be a separate resource that takes a Team as an Input and a member name and then calls the API to do it.

[...]

PR #78, opened 2022-07-26
Return partially created team if adding a member fails

mchaynes:
### Description

Implement partial create failure for Team resource. This fixes an issue where a failure requires a user to go to the UI to recreate a team if they specified a team member that doesn't exist.

Luke phrased what is happening under the hood here better than I can: 
> I believe the "right" way to handle this is the "partial" create failure feature, which allows the provider to indicate that it did create the resource but failed to initialize it. Then in this case the team won't be destroyed, but future changes will be processed as updates to the team which can succeed without a leak.

**Old description**:
~Delete team if adding members fails during creation. We want to delete the team because otherwise, we leave an orphaned team around. The orphaned team will cause a 409 conflict when the user attempts to recreate the team after encountering the error.~

### Related Issues
Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/73


provider/pkg/provider/team_test.go
```diff
@@ -0,0 +1,103 @@
...
+				ReqMethod:  http.MethodDelete,
+			},
+		}
+
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// look up call and increment counter so we can look up the next call
+			wantCall := wantCalls[callNum]
+			callNum++
+
+			// assert that the actual
```
mchaynes, line 65:
```suggestion
			// assert that the actual request matches what we want
```

lukehoban (commented):
I believe the "right" way to handle this is the "partial" create failure feature, which allows the provider to indicate that it did create the resource but failed to initialize it. Then in this case the team won't be destroyed, but future changes will be processed as updates to the team which can succeed without a leak. 

See for example https://github.com/pulumi/pulumi-pulumistack/blob/f1ab9c1333a19732287a00b776ff6fc535ab69ea/provider/pkg/provider/provider.go#L233. 

Cc @stack72 @Frassle @iwahbe  as well for guidance on the right design pattern to use here.

mchaynes:
> 



> I believe the "right" way to handle this is the "partial" create failure feature, which allows the provider to indicate that it did create the resource but failed to initialize it. Then in this case the team won't be destroyed, but future changes will be processed as updates to the team which can succeed without a leak.
> 
> See for example https://github.com/pulumi/pulumi-pulumistack/blob/f1ab9c1333a19732287a00b776ff6fc535ab69ea/provider/pkg/provider/provider.go#L233.

Very interesting! I wasn't aware of this feature. Will use this approach instead.

provider/pkg/provider/team.go
```diff
@@ -195,28 +194,33 @@ func (tr *PulumiServiceTeamResource) Create(req *pulumirpc.CreateRequest) (*pulu
...
-			return nil, err
+			return nil, partialError(*team, err, inProgTeam, inputsTeam)
 		}
+		// if we've successfully added member to team, save them to the state we're going to return
+		// so that a re-run can detect the left over members to add via Update
+		inProgTeam.Members = append(inProgTeam.Members, memberToAdd)
 	}
 
-	outputStore := resource.PropertyMap{}
-	outputStore["__inputs"] = resource.NewObjectProperty(inputs)
```
mchaynes, line 212: For the sake of simplicity, I lifted the input properties from `__inputs` to the top level in the output. 

Initially, I was concerned that lifting this would conflict with resources that were created with an older version of the provider. However, this didn't seem to be an issue, as inside `Update` & `Diff` we reference `inputs` directly.

I tested this by creating a team with the old version of the provider, then upgrading the provider, and then running another update. Diffs and updates worked as expected.

blampe (approved):

Issue #63, opened 2022-05-17
Add additional option to get credentials


Eric-Swiftly:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently there are 3 ways to get credentials to the provider.
set via environment variable
A config file setting
passed as a value while creating a provider.

The first requires an action by users when they checkout the repo.  The second two would require checking in a token to a source repository which is a big no-no.

The users token is already stored in the credentials.json (located in .pulumi/credentials.json in Unix) It would be nice if a 4th option was added to read the credentials from the credentials.json.

mchaynes:
This seems like a reasonable ask to me. The main blocker I see is defining the contract between the pulumi CLI and the pulumi service provider so we can guarantee that this will work as the pulumi CLI is updated. I've filed https://github.com/pulumi/pulumi/issues/9679 to discuss this.

PR #80, opened 2022-08-10
Fetch credentials from disk if not specified via env or config

mchaynes:
## Description

Adds support for fetching pulumi service credentials from disk, instead of requiring it to be set via a config variable or environment variable.

This simplifies the setup and usage of the pulumi service provider. Now, all you need to do is `pulumi login`, like you would normally, and the provider will use those credentials. I've updated the README to reflect this.

Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/63

## Testing
New unit tests, plus manual verification.

### DIFF
--- a/provider/pkg/provider/config.go
+++ b/provider/pkg/provider/config.go
@@ -3,7 +3,6 @@ package provider
 import (
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 )
--- a/provider/pkg/provider/config.go
+++ b/provider/pkg/provider/config.go
@@ -3,14 +3,24 @@ package provider
 import (
 	"fmt"
 	"os"
+	"strings"
+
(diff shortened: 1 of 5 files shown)


provider/pkg/provider/config_test.go
```diff
@@ -0,0 +1,90 @@
...
+			},
+		}
+		gotToken, err := c.getPulumiAccessToken()
+		assert.NoError(t, err)
+		assert.Equal(t, wantToken, *gotToken)
+
+	})
+
+	t.Run("Uses Environment Variable", func(t *testing.T) {
+
```
mchaynes, line 42:
```suggestion
```

provider/pkg/provider/config.go
```diff
@@ -3,8 +3,18 @@ package provider
 import (
 	"fmt"
 	"os"
+	"strings"
```
mchaynes, line 6:
```suggestion
```

djgrove (approved):

PR #81, opened 2022-08-11
Add TeamStackPermission resource

mchaynes:
Adds a resource that provisions TeamStackPermissions to this team

### Reviewers:
You can pretty safely ignore anything in the `sdk/` folder, since that is all generated code.

Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/51

### Testing
* New example that is tested automatically on PR builds
* New unit tests for API calls


blampe:
> You can pretty safely ignore anything in the sdk/ folder, since that is all generated code.

You can automatically hide these from GitHub reviews with gitattributes:
```
$ cat .gitattributes
sdk/* linguist-generated=true
```

mchaynes:
> You can automatically hide these from GitHub reviews with gitattributes

🤯 amazing!

.github/workflows/main.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       fail-fast: true
       matrix:
         dotnetversion: [3.1.301]
-        goversion: [1.17.x]
```
mchaynes, line 107: I needed to update our Go deps because I needed to update our codegen version to use enums.

So I did all that in this PR for simplicity. Happy to break out into a separate PR if that's preferable

blampe (approved):

Issue #85, opened 2022-08-16
Error when refreshing stack that includes a Pulumi webhook


742617000027:
### What happened?

We're using a Pulumi webhook—instantiated with the pulumiservice provider—to monitor activity. It works well when initially created, but upon `pulumi refresh`, the following error occurs (identifying information replaced with `[...]`):

```
pulumiservice:index:Webhook ([...]-pulumi-webhook):
    error: error reading from server: read tcp 127.0.0.1:56677->127.0.0.1:56674: wsarecv: An existing connection was forcibly closed by the remote host.

  pulumi:pulumi:Stack ([...]):
    panic: interface conversion: interface {} is nil, not resource.PropertyMap
    goroutine 98 [running]:
    [github.com/pulumi/pulumi/sdk/v3/go/common/resource.PropertyValue.ObjectValue(...)](http://github.com/pulumi/pulumi/sdk/v3/go/common/resource.PropertyValue.ObjectValue(...))
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.37.2/go/common/resource/properties.go:431
    [github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*PulumiServiceWebhookResource).Diff(0x184f460](http://github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*PulumiServiceWebhookResource).Diff(0x184f460)?, 0xc0000c4000)
        /home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/webhook.go:132 +0x64b
    [github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*pulumiserviceProvider).Diff(0xc00058c3c0](http://github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*pulumiserviceProvider).Diff(0xc00058c3c0), {0xc0001c5880?, 0x101fb55?}, 0x40?)
        /home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/provider.go:181 +0x74
    [github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Diff_Handler.func1({0x18588a8](http://github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Diff_Handler.func1(%7B0x18588a8), 0xc00068c390}, {0x1623ca0?, 0xc0000c4000})
[...]

PR #87, opened 2022-08-17
Fix webhook panic after refresh

mchaynes:
### Description

Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/85

The reason the panic happened was due to a mismatch between the properties produced by `Read` and the properties expected by `Diff`. You can see the difference in output states in more detail below, but the TL;DR is that we were adding an extra object with the key `__inputs` in our output property map upon `Create` and `Update`. `Read` did not stuff values into the `__inputs` object after reading.

In order to simplify things, I lifted every key we previously stored in `__inputs` to the top level, while still handling the old state. So when users upgrade to this new version, on `pulumi up` or `pulumi refresh`, they will see a diff as if we are adding additional output properties, and removing old properties. We don't do this `__inputs` thing in any other providers (See [`pulumi-command`](https://github.com/pulumi/pulumi-command/blob/master/provider/pkg/provider/provider.go#L164) for example)

Additionally, I upgraded examples/go.mod to `go1.19` so that I could upgrade our pinned version of `pulumi/pulumi/sdk` and `pulumi/pulumi/pkg`.

#### Before Refresh:
```json
"inputs": {
    "active": true,
    "displayName": "yaml-webhook",
    "organizationName": "service-provider-test-org",
    "payloadUrl": "https://google.com"
},
"outputs": {
    "__inputs": {
        "active": true,
        "displayName": "yaml-webhook",
        "organizationName": "service-provider-test-org",
        "payloadUrl": "https://google.com"
    },
    "name": "03dbae8b"
},
```
#### After Refresh
```json
"inputs": {
    "__inputs": {
        "active": true,
        "displayName": "yaml-webhook",
        "organizationName": "service-provider-test-org",
        "payloadUrl": "https://google.com"
    },
    "name": "03dbae8b"
},
"outputs": {
    "active": true,
    "displayName": "yaml-webhook",
    "organizationName": "service-provider-test-org",
    "payloadUrl": "https://google.com",
    "secret": ""
},
```

### After This PR
(Still handles old state where `__inputs` was defined)
```json
"inputs": {
    "active": true,
    "displayName": "yaml-webhook",
    "organizationName": "service-provider-test-org",
    "payloadUrl": "https://google.com",
    "secret": ""
},
"outputs": {
    "active": true,
    "displayName": "yaml-webhook",
    "name": "03dbae8b",
    "organizationName": "service-provider-test-org",
    "payloadUrl": "https://google.com"
},
```

### Testing
* Added new test cases to check that refresh works in nodejs and yaml.

##### v0.1.6 --> PR version, with `refresh`
1. Verified installed `v0.1.6` version of the pulumi-service provider.
2. Ran `pulumi up` in `examples/yaml-webhooks`.
3. Ran `pulumi refresh`
4. Upgraded pulumi-service provider to PR version.
5. Ran `pulumi up`
6. Observed `pulumi up` ran successfully and lifted values from `__inputs` to top-level outputs

##### v0.1.6 --> PR version, without `refresh` on v0.1.6
1. Verified installed `v0.1.6` version of the pulumi-service provider.
2. Ran `pulumi up` in `examples/yaml-webhooks`.
4. Upgraded pulumi-service provider to PR version.
5. Ran `pulumi up`
6. Observed `pulumi up` ran successfully and lifted values from `__inputs` to top-level outputs


examples/yaml-webhooks/Pulumi.yaml
```diff
@@ -0,0 +1,16 @@
...
+  webhook:
+    type: pulumiservice:index:Webhook
+    properties:
+      active: true
+      displayName: yaml-webhook
+      organizationName: service-provider-test-org
+      payloadUrl: "https://google.com"
+
+outputs:
+  # export the value of the access token
```
mchaynes, line 15:
```suggestion
  # export the name of the webhook
```

provider/pkg/internal/pulumiapi/webhooks.go
```diff
@@ -42,11 +42,11 @@ type updateWebhookRequest struct {
 	OrganizationName string `json:"organizationName"`
 	DisplayName      string `json:"displayName"`
 	PayloadURL       string `json:"payloadUrl"`
-	Secret           string `json:"secret"`
+	Secret           *string `json:"secret"`
```
blampe, line 45: Do you want omitempty here, too?

blampe (approved):
Ah I figured the `__inputs` was some Pulumi thing, I guess not! Nice cleanup.

provider/pkg/internal/pulumiapi/webhooks.go
mchaynes, line 45:
```suggestion
	Secret           *string `json:"secret,omitempty"`
```

mchaynes: Good catch, yeah.

Issue #86, opened 2022-08-17
Problems updating a stack tag's value


christophermaier:
### What happened?

This is a follow-up to #75; the background and reproduction steps on that issue are still applicable here.

Now, I'm able to change a tag value in the SaaS, but still get empty diffs on subsequent `pulumi up` calls.

Attempting to delete the tag now results in the following error message:
```
error: failed to make request: failed to parse response body, status code 404: invalid character 'p' after top-level value
```

### Steps to reproduce

See #75.

### Expected Behavior

A stack tag should be able to be renamed.

### Actual Behavior

See above.

### Output of `pulumi about`

```
CLI          
Version      3.37.2
Go Version   go1.17.12
Go Compiler  gc

Plugins
NAME           VERSION
pulumiservice  unknown
yaml           unknown

Host     
OS       debian
Version  11.4
Arch     x86_64
```
Though the version of the `pulumiservice` plugin isn't displayed in `pulumi about`, I'm using v0.1.6, based on the contents of `~/.pulumi/plugins`.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

christophermaier:
Thanks, @mchaynes! 🙏

mchaynes:
No problem at all 😄 @christophermaier

PR #88, opened 2022-08-17
Return properties from Update of StackTags

mchaynes:
### Description

Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/86.

We weren't returning any properties from `Update`, which resulted in a diff being produced upon every `pulumi up`. Then, if `pulumi destroy` was run, the empty properties returned from previous `Update` caused the `Delete` call to not have any properties to properly look up the organization or stack name, which meant that the resulting API call failed as well.

### Testing

Modified `TestYamlStackTagsExample` to run another `pulumi up`, this time expecting no changes. This covers the same behavior as the manual test case below.

#### Manual Validation:
(from `examples/yaml-stack-tags` directory)
1. `pulumi up`
2.  modify `value` stack tag property to something else
3. `pulumi up`
4. Run `pulumi up` again. Observed no diff
5. Run `pulumi destroy`. Observed stack tags properly deleted.

### DIFF
--- a/examples/examples_yaml_test.go
+++ b/examples/examples_yaml_test.go
@@ -140,7 +140,7 @@ func TestYamlStackTagsExample(t *testing.T) {
 
 	cwd, _ := os.Getwd()
 
-	os.Setenv("PULUMI_TEST_OWNER", ServiceProviderTestOrg)
+	t.Setenv("PULUMI_TEST_OWNER", ServiceProviderTestOrg)
 	integration.ProgramTest(t, &integration.ProgramTestOptions{
 		Quick:       true,
 		SkipRefresh: true,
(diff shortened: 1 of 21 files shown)


examples/examples_yaml_test.go
```diff
@@ -139,6 +139,8 @@ func TestYamlStackTagsExample(t *testing.T) {
 	tmpdir := writePulumiYaml(t, newProgram)
 
 	cwd, _ := os.Getwd()
+
+	os.Setenv("PULUMI_TEST_OWNER", ServiceProviderTestOrg)
```
blampe, line 143: Handle cleanup:
```suggestion
	t.Setenv("PULUMI_TEST_OWNER", ServiceProviderTestOrg)
```

blampe (approved):
s/os/t otherwise lgtm

PR #89, opened 2022-08-18
Fix YAML Team Stack Permissions example

meagancojocar:


examples/yaml-team-stack-permissions/Pulumi.yaml
```diff
@@ -1,6 +1,6 @@
 name: yaml-team-stack-example
 runtime: yaml
-description: An example of using python to create a team
+description: An example of using yaml to create a team
```
mchaynes, line 3: "An example of using yaml to manage team stack permissions"

mchaynes (changes requested):

mchaynes (approved):

PR #90, opened 2022-08-22
Set test owner to our test organization

mchaynes:
https://github.com/pulumi/pulumi-pulumiservice/pull/88 introduced a change that modified the expected organization for a few tests. I forgot to update `main.yml` as well with that change, so this should fix the build.

Example failing run: https://github.com/pulumi/pulumi-pulumiservice/runs/7907754936?check_suite_focus=true

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,8 @@ env:
   PULUMI_BACKEND_URL: https://api.pulumi-staging.io
   PULUMI_GO_DEP_ROOT: ${{ github.workspace }}/..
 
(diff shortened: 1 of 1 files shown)


blampe (approved):
So it's the release flow that's broken?

mchaynes:
> So it's the release flow that's broken?

Its not the release flow. `release.yml` contains that worfklow. This change fixes the workflow that runs on our `main` branch that tests that `main` is in a good state. We don't have any automation that enforces that `main.yml` is green before we run `release.yml`, but we expect a green build before we run any releases.

komalali (approved):

PR #96, opened 2022-12-10
Ensure forward build compatibility for resource provider

AaronFriel:
See: pulumi/pulumi#11616


iwahbe (approved):
LGTM. 

I'm not sure who we should talk to about failing tests, but they don't look related.

PR #113, opened 2023-03-07
Update dependencies

komalali:
Updates the pulumi dependency to 3.56.0

Finally tracked down the test failures were caused by [a change in pulumi/pulumi](https://github.com/pulumi/pulumi/commit/3b843ffbfc8a344881fda0dfcb6881c9be20220a#diff-58bdc2879a0109885b20b1ac1c52e30423b544b4a0b3e4326c1b79a063c7ce14R382-R384) that requires `PULUMI_TEST_USE_SERVICE` be set (or `RequireService` option be set per test) for `PULUMI_TEST_OWNER` env var to be respected.


komalali:
The test failures are artificial, the actual functionality works just fine. Commented out for now and tracked in https://github.com/pulumi/pulumi-pulumiservice/issues/117

caseyyh (approved):

PR #118, opened 2023-03-20
Add Team/Org access token values and mark as secret

JDTX0:
The current implementation of TeamAccessTokens & OrgAccessTokens have no way to access the value through Pulumi, and the output is also not marked as a secret. 

The raw token is stored in the resource output, which it shouldn't be.

This PR adds the following field to both Team/Org access tokens schema:
```
        "value": {
          "description": "The token's value.",
          "type": "string",
          "secret": true
        }
```

This is to mirror the schema for the standard `pulumiservice:index:AccessToken`

The diffs for everything other than `provider/cmd/pulumi-resource-pulumiservice/schema.json` were generated from `make build`

### DIFF
--- a/provider/cmd/pulumi-resource-pulumiservice/schema.json
+++ b/provider/cmd/pulumi-resource-pulumiservice/schema.json
@@ -156,6 +156,11 @@
         "organizationName": {
           "description": "The organization's name.",
           "type": "string"
+        },
+        "value": {
+          "description": "The token's value.",
+          "type": "string",
+          "secret": true
         }
       },
       "inputProperties": {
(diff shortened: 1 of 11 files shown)


JDTX0:
@jaxxstorm sorry for the mention, but could you take a look at this? it's the only thing holding us back from using team access tokens.

jaxxstorm:
Looks good, sorry I missed this in the initial implementation!

JDTX0:
@jaxxstorm No problem, appreciate you getting it added in!

Can you run the acceptance tests on this PR w/ `/run-acceptance-tests` so auto-merge will continue?

djgrove (approved):
Thanks for your contribution here!

EvanBoyle:
Clean run via https://github.com/pulumi/pulumi-pulumiservice/actions/runs/4494173445 merging.

PR #119, opened 2023-03-23
Fix typo, was passing TeamName instead of Name

JDTX0:
In the `CreateTeamAccessToken` call, it's passing `TeamName` twice, once for the token name, and again for the actual team name.

This leads to issues when you want to create >1 token per team.
`409 API error: Conflict: token name already exists`

The UI also confirmed that the TeamName is being set as the token name.
image

I checked Org access tokens, it doesn't seem to have this error in the code, so just team access tokens have this bug.


JDTX0:
@jaxxstorm @djgrove Can you have a look at this when you get time? We can't make more than 1 access token per team due to this bug.

komalali:
Passing tests: https://github.com/pulumi/pulumi-pulumiservice/actions/runs/4505005078

komalali (approved):

Issue #97, opened 2023-01-29
Support configuring Pulumi Deployments settings for a stack


EvanBoyle:
Related: https://github.com/pulumi/service-requests/issues/169

It would be nice to be able to code review deployment settings for a stack and automatically keep them up to date against a definition checked into source control. The pulumi service provider seems like a natural way to do this without having to create another configuration format.

PR #122, opened 2023-03-28
Deployment settings

komalali:
This PR adds the Pulumi Deployment Settings resource to the provider. The serde code in provider/deployment_settings.go is somewhat unpleasant to behold but I have tested it via automated tests and lots of local testing (especially for the import case). Definitely open to it if there are any quick wins to improve some of this code, but I'd prefer to avoid too much refactoring at the moment and would prefer to get it in the hands of users.

In the future we should definitely consider migrating to the go framework as it will surely cut down much of the boilerplate associated with a hand-rolled provider.

Fixes: https://github.com/pulumi/pulumi-pulumiservice/issues/97


provider/pkg/provider/deployment_settings.go
```diff
@@ -0,0 +1,585 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
+	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
+)
+
+type PulumiServiceDeploymentSettingsInput struct {
+	pulumiapi.DeploymentSettings
+	Stack pulumiapi.StackName
+}
+
+const FixMe = "FIX ME"
```
pgavlin, line 21: Probably want something more direct than `FIX ME` for this.

provider/pkg/provider/deployment_settings.go
```diff
@@ -0,0 +1,585 @@
...
+	replaces := make([]string, len(diffs.ChangedKeys()))
+	for i, k := range diffs.ChangedKeys() {
+		replaces[i] = string(k)
+	}
+
+	return &pulumirpc.DiffResponse{
+		Changes:             pulumirpc.DiffResponse_DIFF_SOME,
+		Replaces:            replaces,
+		DeleteBeforeReplace: true,
+	}, nil
```
pgavlin, line 485: You can use [plugin.NewDetailedDiffFromObjectDiff](https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3@v3.61.1/go/common/resource/plugin#NewDetailedDiffFromObjectDiff) to convert `diff` to a detailed diff that you can return from this method. Personally I'd prefer that to using the deprecated `Replaces` field. You will need to post-process the result to change any diffs to replaces.

provider/pkg/provider/deployment_settings.go
```diff
@@ -0,0 +1,585 @@
...
+
+	dsInput := PulumiServiceDeploymentSettingsInput{
+		Stack:              stack,
+		DeploymentSettings: *settings,
+	}
+
+	properties, err := plugin.MarshalProperties(
+		dsInput.ToPropertyMap(),
+		plugin.MarshalOptions{},
+	)
```
pgavlin, line 529: One sort of unfortunate side-effect of the underlying API is that performing a `refresh` will hand back state that will always cause a diff due to the missing secret values. Not sure that I have a good suggestion for avoiding that.

provider/pkg/provider/deployment_settings.go
```diff
@@ -0,0 +1,585 @@
...
+	}
+	return &pulumirpc.CreateResponse{
+		Id:         path.Join(inputs.Stack.OrgName, inputs.Stack.ProjectName, inputs.Stack.StackName),
+		Properties: req.GetProperties(),
+	}, nil
+}
+
+func (ds *PulumiServiceDeploymentSettingsResource) Update(req *pulumirpc.UpdateRequest) (*pulumirpc.UpdateResponse, error) {
+	// For simplicity, all updates are destructive, so we just call Create.
+	return &pulumirpc.UpdateResponse{}, nil
```
pgavlin, line 580: This should return an error.

pgavlin (changes requested):

provider/pkg/provider/deployment_settings.go
komalali: Do you have a suggestion?

pgavlin: I think anything that indicates that the value needs to be replaced and why would work. Maybe `<value is secret and must be replaced>` or something similar?

provider/pkg/provider/deployment_settings.go
komalali: Any advice on how to go from `plugin.PropertyDiff` to `pulumirpc.PropertyDiff`?

komalali: nvm I think I got it

provider/pkg/provider/deployment_settings.go
komalali: The only thing that comes to mind is by making some affordances in the API to return the secret values. I think for now this is fine, we can document it as a limitation for now.

provider/pkg/provider/deployment_settings.go
komalali: Hm, are you saying you don't want me to use the `Replaces` field? It doesn't mention anywhere that it is deprecated and not using it turns changes into an `update` instead of a `replace` even if every field in the DetailedDiff is updated to have a `Kind` of `Replace`.

pgavlin: > not using it turns changes into an update instead of a replace even if every field in the DetailedDiff is updated to have a Kind of Replace.

Huh, that's surprising. That feels like a bug.

komalali: okay nvm again figured it out :)

pgavlin (approved):

provider/pkg/provider/deployment_settings.go
komalali: Added an issue to track: https://github.com/pulumi/pulumi-pulumiservice/issues/123

Issue #124, opened 2023-04-10
panic: interface conversion: interface {} is nil, not bool


EvanBoyle:
https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/deployment_settings.go#L227

rough guess that this should be `previewPullRequests` instead of `deployPullRequests`

```
    const corpSettings = new pulumiservice.DeploymentSettings("corpSettings", {
        github: {
            deployCommits: true,
            previewPullRequests: true,
            repository: "pulumi/deployment-settings",
        },
        operationContext: {
            oidc: corpAWSOIDCConfig,
            options: {},
        },
        organization: "pulumi",
        project: "deployment-settings",
        sourceContext: {
            git: {
                branch: "refs/heads/main",
            },
        },
        stack: "corp",
    });
```

```
pulumiservice:index:DeploymentSettings (corpSettings):
    error: error reading from server: EOF

  pulumi:pulumi:Stack (deployment-settings-corp):
    panic: interface conversion: interface {} is nil, not bool
    goroutine 43 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/resource.PropertyValue.BoolValue(...)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.56.0/go/common/resource/properties.go:413
    github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.toGitHubConfig(0x1ba0f20?)
    	/home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/deployment_settings.go:227 +0x794
    github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*PulumiServiceDeploymentSettingsResource).ToPulumiServiceDeploymentSettingsInput(0x10f2465?, 0xc000100a46?)
    	/home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/deployment_settings.go:189 +0x214
    github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*PulumiServiceDeploymentSettingsResource).Create(0xc000520090, 0xc000280d70)
    	/home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/deployment_settings.go:573 +0x99
[...]

PR #125, opened 2023-04-10
Fix Panic from marshalling `deployPullRequests` instead of `previewPullRequests`

EvanBoyle:
fixes https://github.com/pulumi/pulumi-pulumiservice/issues/124

Should we add an examples test that covers github integration?


komalali (approved):
I did really want to add a test for the GitHub case but it requires the GitHub app to be installed - I think we might need a new GitHub org to install the app to and I also don't have access to this Pulumi org that is being used for the tests here (it's in staging). All to say, we should definitely add that test but there's some friction there.

EvanBoyle:
Ah, yes. We have seen these struggles with testing against the github app before. Opened https://github.com/pulumi/pulumi-pulumiservice/issues/126 as a follow up.

PR #135, opened 2023-05-13
Support stack webhooks

komalali:
Previously, our support for webhooks in the service provider only covered organization webhooks. This change adds support for stack webhooks as well. 

Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/136


pgavlin (approved):
LGTM!

PR #139, opened 2023-05-18
Fix a bug in import for webhooks

komalali:
Fixes a bug in import for webhooks where we were passing in empty inputs as the properties during an import.


pgavlin (commented):
Are the input and output schemas for web hooks identical?

komalali:
@pgavlin not exactly, output has `name` but input doesn't.

komalali:
@pgavlin fixed!

pgavlin (approved):

PR #141, opened 2023-05-19
Fix diff for Webhook

komalali:


provider/pkg/provider/webhook.go
```diff
@@ -265,42 +265,43 @@ func (wh *PulumiServiceWebhookResource) Diff(req *pulumirpc.DiffRequest) (*pulum
...
-
-				if replace {
-					replaces = append(replaces, key)
-				}
-			}
+	for k, v := range dd {
+		// we don't care about the name property because it is only an output
+		if v.Kind == 2 && k == "name" {
+			continue
+		}
```
komalali, line 287: If I don't do this, the `name` property *always* shows up as a `Delete` and the resource *always* shows an `Update`. I feel like I shouldn't have to do this because it's defined in the schema that `name` is just an `input` and not an `output`. Any ideas what I'm doing wrong here?

pgavlin: I don't think that you're doing anything wrong. I think that this is because `olds` and `news` are shaped differently and the differ is not schema-aware. You could avoid this by preprocessing `news` to add a matching `name` property or pre-processing `olds` to remove the `name` property.

pgavlin (approved):

PR #144, opened 2023-06-03
Fix CI/CD collision with pulumi-command

AaronFriel:
When an already tagged commit is pushed to main, the `main.yml` workflow would run and `pulumictl get version` would return that tag, e.g.: 0.7.2. The goreleaser config would then publish artifacts as `pulumi-command`, resulting in pulumi/pulumi-command#200.

See logs here: https://github.com/pulumi/pulumi-pulumiservice/actions/runs/5082592752/jobs/9132840753#step:9:79

We change two things in this commmit to mitigate conflicts:

1. We change the `main.yml` workflow to exit when the version does not contain a prerelease semver tag of `-alpha`, `-beta`, or `-rc`, addressing a process issue that could result in future conflicts where the release & main workflow attempt to push to the same version.
2. We fix goreleaser prerelease config to publish artifacts as `pulumi-pulumiservice`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -188,6 +188,12 @@ jobs:
         role-to-assume: ${{ secrets.AWS_UPLOAD_ROLE_ARN }}
     - name: Set PreRelease Version
       run: echo "GORELEASER_CURRENT_TAG=v$(pulumictl get version --language generic)" >> $GITHUB_ENV
+    - name: Check PreRelease Version and Exit if not Alpha, Beta, or RC
+      run: |
+        if [[ ! "${GORELEASER_CURRENT_TAG}" =~ -(alpha|beta|rc) ]]; then
+          echo "Error: GORELEASER_CURRENT_TAG does not contain -alpha, -beta, or -rc"
+          exit 1
+        fi
     - name: Run GoReleaser
       uses: goreleaser/goreleaser-action@v2
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -188,6 +188,12 @@ jobs:
         role-to-assume: ${{ secrets.AWS_UPLOAD_ROLE_ARN }}
     - name: Set PreRelease Version
       run: echo "GORELEASER_CURRENT_TAG=v$(pulumictl get version --language generic)" >> $GITHUB_ENV
+    - name: Check PreRelease Version and Exit if not Alpha, Beta, or RC
+      run: |
+        if [[ ! "${GORELEASER_CURRENT_TAG}" =~ -(alpha|beta|rc) ]]; then
+          echo "Error: GORELEASER_CURRENT_TAG does not contain -alpha, -beta, or -rc"
+          exit 1
+        fi
```
blampe, line 196: I'm assuming this is consistent with other workflows?

blampe (approved):

.github/workflows/main.yml
AaronFriel: On the ecosystem & providers team, we limit pushes so that only pull requests can push to main, which prevents us from pushing already tagged commits. (Probably, as I write this I realize it's possible but not obvious how to do that.)

AaronFriel: Looking at audit logs, it looks like this is just defensive coding, happy to remove.

AaronFriel: It looks like the PR was merged, `main.yml` workflow began running, and the release was tagged shortly after; but before the publish *job* ran. When the publish job ran, it picked up that the commit was already tagged 0.7.2.

Issue #151, opened 2023-07-07
GitHub Managed Team creation fails with '400 API error: Bad Request: No TeamID provided'


vschettino-asimov:
### What happened?

Creating a new GitHub Managed team fails with a 400 error: 

```
    error: error creating team '***': failed to create team: 400 API error: Bad Request: No TeamID provided.
```

The [API docs](https://www.pulumi.com/docs/pulumi-cloud/cloud-rest-api/#create-team) do not say anything about TeamID.  

### Expected Behavior

Be able to manage GitHub teams on Pulumi

### Steps to reproduce

Minimal Reproducible Example:

```ts
import * as service from '@pulumi/pulumiservice';

let team = new service.Team(
    'my-team',
    {
    name:  'my-team', // identical to the GitHub Team Name
    organizationName: 'my-org',
    teamType: 'github',
    },
);
```

`pulumi up` will fail with:

```
Diagnostics:
  pulumi:pulumi:Stack (my-stack):
    error: update failed

  pulumiservice:index:Team (my-team):
    error: error creating team 'my-team': failed to create team: 400 API error: Bad Request: No TeamID provided.

Resources:
    2 unchanged

Duration: 4s
```

### Output of `pulumi about`

```
pulumi about
CLI
Version      3.74.0
Go Version   go1.20.5
Go Compiler  gc

Plugins
NAME                     VERSION
datadog                  4.19.1
gcp                      6.59.0
kubernetes               3.30.1
kubernetes-cert-manager  0.0.1
nodejs                   unknown
pulumiservice            0.9.0

Host
OS       darwin
Version  13.4.1
Arch     arm64

This project is written in nodejs: executable='/Users/vschettino/.nvm/versions/node/v16.14.2/bin/node' version='v16.14.2'

Dependencies:
NAME                              VERSION
@pulumi/pulumi                    3.74.0
@pulumi/pulumiservice             0.9.0
@asimovbio/pulumi-gcp-components  1.9.8
typescript                        5.1.6
@types/node                       18.16.19


Pulumi locates its logs in /var/folders/0x/7l81mby95f9_b26gzzrqqfdw0000gn/T/ by default
```

### Additional context

- This is a blocker because `pulumiservice.Team.get()` is not implemented yet.
- Managing `teamType: pulumi` works just fine.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

[...]

PR #153, opened 2023-07-17
Expose GitHub team ID

blampe:
The team creation API for GitHub teams only requires a `githubTeamID` parameter, which it uses to fetch all the other details like team name, description, members, etc.

I implemented `Read` while I was in here, because it was surprising to see this break `refresh` due to a not-implemented error. Also implemented `Check` because `name` and `githubTeamId` are only required for certain team types.

Tested and confirmed working by with the following program:

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as service from "@pulumi/pulumiservice";

const team = new service.Team("github-team", {
  organizationName: "pulumi",
  teamType: "github",
  githubTeamId: 7475955,
});
``` 

```bash
❯ pulumi up -s pulumi/test
Previewing update (pulumi/test)

View in Browser (Ctrl+O): https://.../pulumi/pulumi-service-teams-example-ts/test/previews/3dd36e22-e7d7-4b5d-82d0-2bbd8d61f86e

     Type                         Name                                  Plan
     pulumi:pulumi:Stack          pulumi-service-teams-example-ts-test
 +   └─ pulumiservice:index:Team  github-team                           create


Resources:
    + 1 to create
    1 unchanged

Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:test::pulumi-service-teams-example-ts::pulumi:pulumi:Stack::pulumi-service-teams-example-ts-test]
    + pulumiservice:index:Team: (create)
        [urn=urn:pulumi:test::pulumi-service-teams-example-ts::pulumiservice:index:Team::github-team]
        [provider=urn:pulumi:test::pulumi-service-teams-example-ts::pulumi:providers:pulumiservice::default_0_9_1_alpha_1689717595_28bbd1fc_dirty::7a11ee03-f53f-4ebc-8c9f-a09cfe4ad07a]
        githubTeamId    : 7475955
        organizationName: "pulumi"
        teamType        : "github"

Do you want to perform this update? yes
Updating (pulumi/test)

View in Browser (Ctrl+O): https://.../pulumi/pulumi-service-teams-example-ts/test/updates/5

     Type                         Name                                  Status
     pulumi:pulumi:Stack          pulumi-service-teams-example-ts-test
 +   └─ pulumiservice:index:Team  github-team                           created (0.94s)


Resources:
    + 1 created
    1 unchanged

Duration: 3s
```

Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/151.

### DIFF
--- a/provider/cmd/pulumi-resource-pulumiservice/schema.json
+++ b/provider/cmd/pulumi-resource-pulumiservice/schema.json
@@ -475,6 +475,10 @@
         "organizationName": {
           "description": "The organization's name.",
           "type": "string"
+        },
+        "githubTeamID": {
+            "description": "The GitHub ID of the team to mirror. This is the only required parameter when creating a GitHub team -- all other parameters are taken from GitHub directly. Must be in the same GitHub organization that the Pulumi org is backed by.",
+            "type": "number"
         }
       },
       "inputProperties": {
--- a/provider/cmd/pulumi-resource-pulumiservice/schema.json
+++ b/provider/cmd/pulumi-resource-pulumiservice/schema.json
@@ -477,7 +477,7 @@
           "type": "string"
         },
         "githubTeamId": {
-            "description": "The GitHub ID of the team to mirror. This is the only required parameter when creating a GitHub team -- all other parameters are taken from GitHub directly. Must be in the same GitHub organization that the Pulumi org is backed by.",
+            "description": "The GitHub ID of the team to mirror. Must be in the same GitHub organization that the Pulumi org is backed by.",
             "type": "number"
         }
       },
@@ -510,7 +510,7 @@
           "type": "string"
         },
         "githubTeamId": {
-            "description": "The GitHub ID of the team to mirror. This is the only required parameter when creating a GitHub team -- all other parameters are taken from GitHub directly. Must be in the same GitHub organization that the Pulumi org is backed by.",
+            "description": "The GitHub ID of the team to mirror. Must be in the same GitHub organization that the Pulumi org is backed by.",
             "type": "number"
         }
       },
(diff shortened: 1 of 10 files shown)


provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -475,6 +475,10 @@
         "organizationName": {
           "description": "The organization's name.",
           "type": "string"
+        },
+        "githubTeamID": {
+            "description": "The GitHub ID of the team to mirror. This is the only required parameter when creating a GitHub team -- all other parameters are taken from GitHub directly. Must be in the same GitHub organization that the Pulumi org is backed by.",
+            "type": "number"
```
lukehoban, line 481: This needs to also be listed as an input property, right?

The schema still marks the other input properties as required, do those need to be made optional now?

lukehoban (commented):
I believe some additional changes to the schema will be needed. Also would be great to add test coverage for this - though not sure how easy it will be to automate that end-to-end. At the very least - a test that runs a preview to ensure this is supported correctly in the generated SDK.

provider/cmd/pulumi-resource-pulumiservice/schema.json
blampe: > The schema still marks the other input properties as required, do those need to be made optional now?

Updated so `name` is now optional, org and type are still needed to construct relevant URLs.

pgavlin: Do we validate the various combinations in `Check`?

pgavlin (approved):
LGTM, just one question about `Check`

provider/cmd/pulumi-resource-pulumiservice/schema.json
pgavlin: > Most of this provider's checks seem to no-op like this

:(

>  If it's worth doing while we're in here I can implement it.

I think that it's worth doing in this case b/c there are different combinations of valid inputs depending on the team type.

Issue #146, opened 2023-06-08
Refresh and destroy on `DeploymentSettings` fails with 404 if there is an out of band deletion


EvanBoyle:
Steps to repro:

1. create deployment settings from console
2. import deployment settings into program `pulumi import pulumiservice:index:DeploymentSettings deploymentSettings pulumi/test/bugrepro`
3. clear deployment settings on the stack via the console
4. run `pulumi refresh --skip-preview` and see the program fail due to 404, same behavior if you try to destroy.

To remediate, you have to export the state, delete the resource, import state, and only then are you unblocked to run an update and recreate the DeploymentSettings resource.

```
➜  test pulumi refresh --skip-preview
Refreshing (pulumi/bugrepro)

View in Browser (Ctrl+O): https://app.pulumi.com/pulumi/test/bugrepro/updates/2

     Type                                       Name                Status                    Info
     pulumi:pulumi:Stack                        test-bugrepro       **failed**                1 error
 ~   └─ pulumiservice:index:DeploymentSettings  deploymentSettings  **refreshing failed**     1 error


Diagnostics:
  pulumi:pulumi:Stack (test-bugrepro):
    error: update failed

  pulumiservice:index:DeploymentSettings (deploymentSettings):
    error: deployment settings for stack (pulumi/test/bugrepro) not found

Resources:
    1 unchanged

Duration: 1s
```

PR #154, opened 2023-08-01
fix: Refresh and destroy on resources with an out of band deletion 

glena:
DeploymentSettings, Team and Webhook api client now returns nil on 404 responses so the provider treats it accordingly and recognize them as deleted resources

Fix https://github.com/pulumi/pulumi-pulumiservice/issues/146

image

image

### DIFF
--- a/provider/pkg/provider/deployment_settings.go
+++ b/provider/pkg/provider/deployment_settings.go
@@ -3,6 +3,7 @@ package provider
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"path"
 
 	pbempty "github.com/golang/protobuf/ptypes/empty"
@@ -529,7 +530,15 @@ func (ds *PulumiServiceDeploymentSettingsResource) Read(req *pulumirpc.ReadReque
 	}
 	settings, err := ds.client.GetDeploymentSettings(ctx, stack)
 	if err != nil {
+		statusCode := pulumiapi.GetErrorStatusCode(err)
+		if statusCode == http.StatusNotFound {
+			// deleteResponse causes the resource to be deleted from the state.
+			var deleteResponse = &pulumirpc.ReadResponse{Id: "", Properties: nil}
+			// If it's a 404 error, this resource was probably deleted.
+			return deleteResponse, nil
+		}
 		return nil, err
+
 	}
 
 	dsInput := PulumiServiceDeploymentSettingsInput{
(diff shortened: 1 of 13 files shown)


pgavlin (approved):

provider/pkg/provider/deployment_settings.go
```diff
@@ -529,7 +530,15 @@ func (ds *PulumiServiceDeploymentSettingsResource) Read(req *pulumirpc.ReadReque
 	}
 	settings, err := ds.client.GetDeploymentSettings(ctx, stack)
 	if err != nil {
+		statusCode := pulumiapi.GetErrorStatusCode(err)
+		if statusCode == http.StatusNotFound {
+			// deleteResponse causes the resource to be deleted from the state.
+			var deleteResponse = &pulumirpc.ReadResponse{Id: "", Properties: nil}
+			// If it's a 404 error, this resource was probably deleted.
+			return deleteResponse, nil
+		}
```
blampe, line 539: This fixes the problem for DeploymentSettings, but it occurs to me that all of our other resources probably have the same bug. Can we do this check one layer up in the provider instead of the resource? _I think_ this would cover it:

https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/provider.go#L205-L209

Aside: several resources don't implement `Read` (ugh) but we don't need to fix that here. It would be so nice to some day generate this provider from an OpenAPI spec!

glena: I will take a look tomorrow morning about that, thx 🙇

glena: I just found teams is handling the "not found" in the way I mention yesterday to you, returning nil instead of propagating the error, I will change to handle it in the same way https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/team.go#L192-L194 (unless the api is returning a 404 too and it is just returning an error? https://github.com/pulumi/pulumi-pulumiservice/blob/master/provider/pkg/internal/pulumiapi/teams.go#L108-L112)

For webhooks, it is weird, I guess it is returning a 404 too? I will ping you to see how can I actually test it (I dont want to use pulumi org and my personal one does not have it enabled https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/webhook.go#L374-L419)

glena: Confirmed, teams and webhooks suffer from the same thing
image

image

I will add those 2 tomorrow morning and ask you again for review, thx for bringing it up

glena: image


image

blampe (approved):
Amazing!

PR #162, opened 2023-08-23
Add support for new skipIntermediateDeployments setting

glena:
Adds support to the new `skipIntermediateDeployments` on `DeploymentSettings`.

Fix https://github.com/pulumi/pulumi-cloud-requests/issues/136


provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -311,6 +317,10 @@
           "type": "boolean",
           "description": "Skip the default dependency installation step - use this to customize the dependency installation (e.g. if using yarn or poetry)"
         },
+        "skipIntermediateDeployments": {
+          "type": "boolean",
+          "description": "Skip duplicated queued operations (it will only execute the last deployment of the same type)"
+        },
```
blampe, line 323: Whatever we end up calling this feature, we should try to use a consistent name. I suggested calling it "skip pending" in your other PR mostly because that's what the user's issue called it.

@meagancojocar what should we call this? "skip duplicated...", "skip intermediate...", "skip pending...", etc.? I don't have a strong opinion, I just want to make sure we use language consistent with whatever we're advertising this as.

blampe (commented):
The change is fine, let's just nail down the naming for consistency.

provider/cmd/pulumi-resource-pulumiservice/schema.json
glena: makes sense

meagancojocar: I'm good with skipIntermediateDeployments. Let's stay away from pending as we use that term for something else in the CLI, duplicated isn't exactly true as they can have different commits, which leaves intermediate as the most descriptive I think. Agree with Bryce that we should be consistent.

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -311,6 +317,10 @@
           "type": "boolean",
           "description": "Skip the default dependency installation step - use this to customize the dependency installation (e.g. if using yarn or poetry)"
         },
+        "skipIntermediateDeployments": {
+          "type": "boolean",
+          "description": "Skip pending deployments (Consolidate multiple deployments of the same type into one deployment)"
```
meagancojocar, line 322:
```suggestion
          "description": "Skip intermediate deployments (Consolidate multiple deployments of the same type into one deployment)"
```

meagancojocar (commented):
LGTM 👍

provider/cmd/pulumi-resource-pulumiservice/schema.json
glena: will change that everywhere

komalali, line 322: Maybe something like "Only run the latest deployment if multiple of the same type are queued"

blampe (approved):

PR #168, opened 2023-09-12
Support org admin tokens.

pgavlin:
- Add an `admin` inout property to OrgAccessToken
- Fix existing output properties for OrgAccessToken

### DIFF
--- a/provider/pkg/provider/org_access_token.go
+++ b/provider/pkg/provider/org_access_token.go
@@ -20,6 +20,7 @@ type PulumiServiceOrgAccessTokenInput struct {
 	OrgName     string
 	Description string
 	Name        string
+	Admin       bool
 }
 
 func (i *PulumiServiceOrgAccessTokenInput) ToPropertyMap() resource.PropertyMap {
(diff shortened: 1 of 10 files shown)


provider/pkg/provider/org_access_token.go
```diff
@@ -70,14 +76,17 @@ func (ot *PulumiServiceOrgAccessTokenResource) Diff(req *pulumirpc.DiffRequest)
...
 	}
 
-	changes := pulumirpc.DiffResponse_DIFF_NONE
+	changes, replaces := pulumirpc.DiffResponse_DIFF_NONE, []string(nil)
 	if diffs.Changed("description") {
-		changes = pulumirpc.DiffResponse_DIFF_SOME
+		changes, replaces = pulumirpc.DiffResponse_DIFF_SOME, append(replaces, "description")
+	}
+	if diffs.Changed("admin") {
+		changes, replaces = pulumirpc.DiffResponse_DIFF_SOME, append(replaces, "admin")
```
komalali, line 84: Should we update this to use `NewDetailedDiffFromObjectDiff` like [here](https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/deployment_settings.go#L490)?

komalali (approved):

provider/pkg/provider/org_access_token.go
pgavlin: Yes, but I'd suggest doing that in a followup.

PR #193, opened 2023-11-20
Update dependencies

komalali:


sdk/nodejs/scripts/install-pulumi-plugin.js
```diff
@@ -1,26 +0,0 @@
-"use strict";
```
blampe, line 1: Is deleting this file expected?

blampe (approved):

sdk/nodejs/scripts/install-pulumi-plugin.js
komalali: Yeah, these were dropped in an earlier version of p/p: 
PR in p/p: https://github.com/pulumi/pulumi/pull/13800
Update in random: https://github.com/pulumi/pulumi-random/commit/5a62a9ba0de746cebe872f22d429ab6d85e48706#diff-742c39e34ce26d6cf7e23d3ec1afccfe491e9e33afdb8514bcb52a30d57ee703

Issue #199, opened 2023-11-21
In `Team` resource, `name` is optional but resource creation fails if unspecified.


aureq:
### What happened?

With the program below, I attempt to create a Pulumi `Team` in my org. The `.name` property if of type `name?: pulumi.Input<string>;` (https://github.com/pulumi/pulumi-pulumiservice/blob/main/sdk/nodejs/team.ts#L124-L127). However, if the `.name` is unspecified, the resource creation fails with the error below.

```
error: pulumiservice:index:Team resource '[coolspot]-team': property name value {<nil>} has a problem: pulumi teams require a name
```

If the name is missing, then the Pulumi Resource name should be used instead as the base for the actual team name.

### Example

```ts
import * as pulumi from "@pulumi/pulumi";
import * as pulumiservice from "@pulumi/pulumiservice";
import * as random from "@pulumi/random";

export = async () => {

    const orgName = pulumi.getOrganization();
    // const teamName = "team-1";
    const pipelineName = "coolspot";

    const tokenSuffix = new random.RandomString(`[${pipelineName}]-pulumi-token-suffix`, {
        length: 7,
        lower: true,
        minLower: 3,
        numeric: true,
        minNumeric: 4,
        special: false,
        upper: false,
    });

    const pulumiTeam = new pulumiservice.Team(`[${pipelineName}]-team`, {
        // name: pulumi.interpolate`${pipelineName}-${tokenSuffix.result}`,
        organizationName: orgName,
        teamType: "pulumi",
    });
}
```

### Output of `pulumi about`

```
CLI          
Version      3.93.0
Go Version   go1.21.3
Go Compiler  gc

Plugins
NAME           VERSION
nodejs         unknown
pulumiservice  0.16.0
random         4.13.2

Host     
OS       debian
Version  12.2
Arch     x86_64

This project is written in nodejs: executable='/usr/local/bin/node' version='v20.9.0'

Current Stack: menfin/pulumiservice-ts/dev

Found no resources associated with dev

Found no pending operations associated with dev

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/aureq
User           aureq
Organizations  aureq, team-ce, menfin, menfin-team, demo, pulumi
Token type     personal

Dependencies:
NAME                   VERSION
@pulumi/pulumi         3.78.0
@pulumi/pulumiservice  0.16.0
@pulumi/random         4.13.2
@types/node            16.18.40

Pulumi locates its logs in /tmp by default
```

[...]

PR #203, opened 2023-11-21
Slightly improve error messages

lukehoban:
The error messages can be slightly more explicit that 'name' is required only if the teamType is "pulumi".

Helps with #199.


aureq (approved):
that looks great.

komalali (approved):

PR #240, opened 2024-04-04
Remove the need for cmd/pulumi-gen-pulumiservice

iwahbe:
We don't need to maintain our own gen code, we can use the pulumi binary for that. This makes it important that we use the right version of the pulumi binary, so we use the Makefile to ensure the version of pulumi used matches the version of pulumi we depend on in `provider/go.mod`.

We use a similar pattern in other native providers.

I'm tagging @pierskarsenbarg and @komalali since they were the suggested reviewers. If I should tag someone else here, please let me know.


komalali:
Any thoughts on what the do about the failing test here @iwahbe? I had to add those empty stack yaml files in my last update of pu/pu because of this change https://github.com/pulumi/pulumi/pull/14695

With your change the stack is named dynamically...

iwahbe:
In the mean time, we can work around the bug by ensuring that the files it wants are created. That is what https://github.com/pulumi/pulumi-pulumiservice/pull/240/commits/671af082c9ec0b6ec95a9b0eff86ba83da5c2405 does.

komalali (approved):
Amazing! Thank you for all your effort here!

PR #244, opened 2024-04-12
Adding delete after destroy into deployment settings

IaroslavTitov:
### Summary
- Added the parameter into schema, generated SDKs
- Added provider logic to use parameter from schema when calling pulumi cloud

### Testing
- imported generated .Net SDK into local project
- verified in console that `pulumi up` created DeploymentSettings with the `deleteAfterDestroy` flag as expected


cleverguy25:
@IaroslavTitov this looks good to me, but would defer to @komalali

komalali:
@cleverguy25 yep I chatted with @IaroslavTitov already, we just need to add a changelog entry and it should be good to go.

IaroslavTitov:
Just added a pending changelog

komalali (approved):

PR #247, opened 2024-04-18
Adding support for DeploymentSchedule resource

IaroslavTitov:
### Summary
- Adding deployment schedules into schema
- Adding ScheduleClient to call schedule APIs
- Adding pulumi resource DeploymentSchedule

### Notes
- Changing any of the stack path variables will trigger a Replace
- Other changes will trigger an Update

### Testing
- Installed locally and tested using DotNet SDK
```
    var schedule = new DeploymentSchedule(
        "deployment Schedule",
        new DeploymentScheduleArgs{
            Project = "PulumiDotnet",
            Stack = "SdkTest3",
            ScheduleCron = "0 0 * * 0",
            PulumiOperation = PulumiOperation.UPDATE
        }
    );
```
The outputs populate correctly, the schedules show up in the console.


provider/pkg/provider/deployment_schedules.go
```diff
@@ -0,0 +1,345 @@
...
+			Property: "projectName",
+		})
+	}
+
+	if inputMap["timestamp"].HasValue() && inputMap["timestamp"].IsString() {
+		_, err := time.Parse(time.RFC3339, inputMap["timestamp"].StringValue())
+		if err != nil {
+			failures = append(failures, &pulumirpc.CheckFailure{
+				Reason:   fmt.Sprintf("timestamp failed to parse due to: %s", err),
+				Property: "projectName",
```
komalali, line 262: Should this be timestamp

provider/pkg/provider/deployment_schedules.go
```diff
@@ -0,0 +1,345 @@
...
+				Property: string(p),
+			})
+		}
+	}
+
+	if (inputMap["scheduleCron"].HasValue() && inputMap["timestamp"].HasValue()) ||
+		(!inputMap["scheduleCron"].HasValue() && !inputMap["timestamp"].HasValue()) {
+		failures = append(failures, &pulumirpc.CheckFailure{
+			Reason:   "One of scheduleCron and timestamp must be specified but not both",
+			Property: "projectName",
```
komalali, line 253: This should be one of scheduleCron or timestamp - I think?

komalali (approved):
Looks great! Just a couple of small comments, feel free to merge after you fix them up.

We should also follow up with an example in /examples but can leave that for later.

provider/pkg/provider/deployment_schedules.go
IaroslavTitov: I did call the external property timestamp to make distinction easier. It is the same way in TTL, but not in raw API. (so logging timestamp to align with the name)

Totally open to renaming the property to `scheduleOnce` if that'd be better, I just think similar names are confusing + in TTL has timestamp.

provider/pkg/provider/deployment_schedules.go
IaroslavTitov: Ah, agree "and" is confusing here, I'll change to "or"

provider/pkg/provider/deployment_schedules.go
komalali: Oh I just meant the Property looks like a copy paste error
```suggestion
				Property: "timestamp",
```

provider/pkg/provider/deployment_schedules.go
komalali: Again, was referring to the Property
```suggestion
			Property: "scheduleCron",
```

provider/pkg/provider/deployment_schedules.go
IaroslavTitov: Oh crap, I didn't even notice that lol. Yes, def copypaste error, fixing

provider/pkg/provider/deployment_schedules.go
IaroslavTitov: Totally flew over my head, fixing

PR #253, opened 2024-04-23
Fixing main.yml workflow

IaroslavTitov:
Removing the second file that no longer exists, as @iwahbe suggested. 

@komalali , is there any way to run this before merging in?


komalali (approved):

komalali:
> @komalali , is there any way to run this before merging in?

You could potentially change the workflow to run on this branch too. But I think it's fine to just merge and have it run on merge.

PR #254, opened 2024-04-23
Add new drift events to webhook resource

komalali:


IaroslavTitov (approved):
I was just about to make codefreeze PR, perfect timing!

QQ: are `value` field magic strings that match code in Pulumi Cloud? I'm guessing so, just want to make they match, or if you tested these?

komalali:
@IaroslavTitov yeah they match here: https://github.com/pulumi/pulumi-service/blob/6eed4db07aa3d861941673cbb69f8ea1599161e1/cmd/service/model/webhooks.go#L60-L64

Issue #123, opened 2023-04-07
Refresh on Deployment Settings always shows a diff if there are secrets


komalali:
The underlying deployment settings API does not return any secret values on `GET`. As a result, on `refresh`, there is always a diff  as the secret values are replaced by the string `secret`. Subsequently, an `update` also shows a diff.

To work through this, we will likely need to implement an endpoint for deployment settings that returns the secret values.

IaroslavTitov:
Investigating into this using this example resource:
```
var settings = new DeploymentSettings(
        "My Settings",
        new DeploymentSettingsArgs{
            Organization = "IaroslavTitov",
            Project = "PulumiDotnet",
            Stack = "SdkTest4",
            SourceContext = new DeploymentSettingsSourceContextArgs{
                Git = new DeploymentSettingsGitSourceArgs{
                    RepoUrl = "https://fake-test.com",
                    Branch = "test",
                    GitAuth = new DeploymentSettingsGitSourceGitAuthArgs {
                        BasicAuth = new DeploymentSettingsGitAuthBasicAuthArgs{
                            Username = "IaroTest",
                            Password = "IaroTest"
                        }
                    }
                }
            }
        }
    );
   ```
   
   Git login into are secrets.
   Pulumi up succeeds, and I can see secrets in the console like so:
   ```
"password": {
          "4dabf18193072939515e22adb298388d": "1b47061264138c4ac30d75fd1eb44270",
          "ciphertext": "AAABAAT6I4md3+xjlqIQf1Lntt+R9PomYXFZdg2RSViuWP2CWEw3k5QN"
        },
        "username": {
          "4dabf18193072939515e22adb298388d": "1b47061264138c4ac30d75fd1eb44270",
          "ciphertext": "AAABABfTbRhhdObvZTKqcY0wxObHcONld3QOXZz1GzdhNASu401yqEdi"
        }
```

When I do pulumi refresh, it wants to update agentPooId and sourceContext, and result looks like this

```
      "basicAuth": {
        "password": {},
        "username": {}
      }
```

So Read operation breaks these secrets, which is why pulumi up shows diff and actually fails after that, failing to read empty object where it expected a string or a secret.

IaroslavTitov:
Had a meeting to discuss this issue, review design doc and agreed on a gameplan described in the doc. Will proceed with implementation.

IaroslavTitov:
[...]

PR #259, opened 2024-04-26
Made secret values in Deployment Settings actually work, save and refresh from source of truth

IaroslavTitov:
### Summary
- Adding ability for client to call GET endpoints with query params (just adding to url didn't work as it got encoded and ? was turned into unicode)
- Adding logic to retrieve deployment settings with secrets on Read operation
- Adjusted logic to properly compare actual values, but only show encrypted values in console
- Other adjusted to prevent comparison of empty fields causing fake drift
- All deploymentSettings's env vars created using this provider will always be secret. I figured adding ability to pick secret or not is out of scope, I can make a followup PR, but schema will neeed to change for that.

### Issues
- [#123](https://github.com/pulumi/pulumi-pulumiservice/issues/123)

### Testing
- tested locally using devstack + DotNet SDK
```
    var settings = new DeploymentSettings(
        "My Settings",
        new DeploymentSettingsArgs{
            Project = "PulumiDotnet",
            Stack = "SdkTest4",
            SourceContext = new DeploymentSettingsSourceContextArgs{
                Git = new DeploymentSettingsGitSourceArgs{
                    RepoUrl = "https://fake-test.com",
                    Branch = "test",
                    GitAuth = new DeploymentSettingsGitSourceGitAuthArgs {
                        // BasicAuth = new DeploymentSettingsGitAuthBasicAuthArgs{
                        //     Username = "IaroTest",
                        //     Password = "IaroPass"
                        // },
                        SshAuth = new DeploymentSettingsGitAuthSSHAuthArgs{
                            SshPrivateKey = "IaroSshKey",
                            Password = "IaroSshPass"
                        }
                    }
                }
            },
            OperationContext = new DeploymentSettingsOperationContextArgs{
                EnvironmentVariables = new InputMap<string>
                {
                    { "IaroEnvVarKey", "IaroEnvVarSecret" }
                }
            }
        }
    );
```

Results in values in the resource explorer like so:
```
{
  "git": {
    "branch": "test",
    "gitAuth": {
      "sshAuth": {
        "password": {
          "ciphertext": "AAABAEsoaNBwL887is1BhzC9c/9JCiErA3v8T0K9pu/ta8Bsk93WXHkuPYGS"
        },
        "sshPrivateKey": {
          "ciphertext": "AAABALukLwq5StYjQQugxJwlS3bXLm55DSrUgVI58U0FzLpztzPDoXpPiF4="
        }
      }
    },
    "repoUrl": "https://fake-test.com"
  }
}
```

but refresh now shows no change UNLESS there was actual change.

### DIFF
--- a/provider/pkg/provider/deployment_settings.go
+++ b/provider/pkg/provider/deployment_settings.go
@@ -19,14 +19,14 @@ type PulumiServiceDeploymentSettingsInput struct {
 	Stack pulumiapi.StackName
 }
 
-const FixMe = "<value is secret and must be replaced>"
-
 func (ds *PulumiServiceDeploymentSettingsInput) ToPropertyMap() resource.PropertyMap {
 	pm := resource.PropertyMap{}
 	pm["organization"] = resource.NewPropertyValue(ds.Stack.OrgName)
 	pm["project"] = resource.NewPropertyValue(ds.Stack.ProjectName)
 	pm["stack"] = resource.NewPropertyValue(ds.Stack.StackName)
-	pm["agentPoolId"] = resource.NewPropertyValue(ds.AgentPoolId)
+	if ds.AgentPoolId != "" {
+		pm["agentPoolId"] = resource.NewPropertyValue(ds.AgentPoolId)
+	}
 
 	if ds.SourceContext != nil {
 		scMap := resource.PropertyMap{}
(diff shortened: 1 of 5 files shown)


provider/pkg/provider/deployment_settings.go
```diff
@@ -349,11 +351,8 @@ func toOperationContext(inputMap resource.PropertyMap) *pulumiapi.OperationConte
 		evInput := ocInput["environmentVariables"].ObjectValue()
 
 		for k, v := range evInput {
-			if v.IsSecret() {
-				ev[string(k)] = apitype.SecretValue{Secret: true, Value: v.SecretValue().Element.StringValue()}
-			} else {
-				ev[string(k)] = apitype.SecretValue{Secret: false, Value: v.StringValue()}
-			}
+			// This hardcodes all env vars created via provider as secret. Schema change would be needed for optional secrecy
+			ev[string(k)] = apitype.SecretValue{Secret: true, Value: v.StringValue()}
```
komalali, line 355: Why is this change necessary? Previously if the value passed in was a pulumi secret we would set it as a secret. Otherwise it would not be marked secret. Looks like we changed this to mark everything as secret, but it's not clear why.

IaroslavTitov:
PR checks are failing, so I made https://github.com/pulumi/pulumi-pulumiservice/pull/261 to fix the wf

provider/pkg/provider/deployment_settings.go
IaroslavTitov: This change was not strictly necessary, but the way things were there was no way to make a secret env var. (Correct me if I'm wrong, env var map was just a map of string to string). 

So I opted for secret always vs non-secret always. As I mentioned in description, it would be great to come back and change the schema and actually allow both secret and non-secret env vars.

IaroslavTitov:
Also this PR is blocked until [this PR](https://github.com/pulumi/pulumi-service/pull/19463) is in

provider/pkg/provider/deployment_settings.go
komalali: Yeah I think you're mistaken. With the current implementation, if you mark something as secret the pulumi way (pulumi.Secret or config.requireSecret) it will set the the env var as secret. It's not in the schema but we do already respect secretness. Recommend reading this doc and trying it out with this change reverted. https://www.pulumi.com/docs/concepts/secrets/

komalali: This test has some examples: https://github.com/pulumi/pulumi-pulumiservice/blob/main/examples/ts-deployment-settings/index.ts

IaroslavTitov: Just tried it out, `Output<String>` is black magic 🤯 

I will need to adjust parsing in `toOperationContext`, because adding secrets turns it into below struct, but other than that it should work I think. 
```
{
	struct_value: {
		fields:{
			key:"environmentVariables"
			value: {
				struct_value:{
					fields:{
						key:"4dabf18193072939515e22adb298388d"
						value:{
							string_value:"1b47061264138c4ac30d75fd1eb44270"
						}
					}
					fields:{
						key:"value"
						value:{
							struct_value:{
								fields:{
									key:"IaroEnvVarKey"
									value:{
										string_value:"[secret]"
									}
								}
							}
						}
					}
				}
			}
		}
	}
} 
```

I will figure out the parsing and update PR, likely next week.

IaroslavTitov:
Closing this outdated PR, finally implemented this fix in https://github.com/pulumi/pulumi-pulumiservice/pull/320

Issue #257, opened 2024-04-24
TtlSchedule and DriftSchedule - pulumi refresh loses schedule ID so update adds schedule


MitchellGerdisch:
### What happened?

I have a stack with a TtlSchedule resource and DriftSchedule resource.
If I run `pulumi refresh -y` the schedules' ids are removed from state.
So, if change the resource (say modify the time or cron) a subsequent `pulumi up` creates additional schedules.

### Example

- Use your favorite pulumi (python) program/stack with deployments configured and add the code below.
- `pulumi up -y`
- See the two schedules that were created.
- `pulumi refresh -y`
   - Note that schedule Ids were removed from state
- Modify the two schedule resources by modifying the time or cron properties
- `pulumi up -y`
- In Pulumi Cloud UI see that there are now two of each type of schedule instead of just modifying the original schedules.

```
org = pulumi.get_organization()
project = pulumi.get_project()
stack = pulumi.get_stack()

ttlSchedule = ps.TtlSchedule("ttlsched", 
                             organization=org,
                             project=project,
                             stack=stack,
                             timestamp="2024-07-01T00:00:00Z",
                             delete_after_destroy=False)

driftSchedule = ps.DriftSchedule("driftsched", 
                             organization=org,
                             project=project,
                             stack=stack,
                             schedule_cron="0 * * * *") 
```

### Output of `pulumi about`

CLI          
Version      3.113.3
Go Version   go1.22.2
Go Compiler  gc

Plugins
KIND      NAME           VERSION
resource  aws            6.32.0
resource  pulumiservice  0.20.0
language  python         unknown
resource  stackmgmt      3.1.0

Host     
OS       darwin
Version  13.6.6
Arch     x86_64

This project is written in python: version='3.10.11'

Backend        
Name           pulumi.com

Dependencies:
NAME              VERSION
pequod-stackmgmt  3.1.0
pip               23.0.1
pulumi_aws        6.32.0
setuptools        65.5.0

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

MitchellGerdisch:
This also makes it impossible to destroy the stack without state surgery due to this error:
`    error: failed to delete schedule with scheduleId  : failed to parse response body, status code 404: invalid character 'p' after top-level value`

IaroslavTitov:
[...]

PR #260, opened 2024-04-26
Fixing refresh breaking drift and ttl schedule

IaroslavTitov:
### Summary
- Fixing 2 bugs:
- [First](https://github.com/pulumi/pulumi-pulumiservice/issues/257) is me forgetting to add scheduleId to output properties on Read
- Second one I discovered while working on the first one, when booleans are not provided, they are still saved to output state, which causes eternal diff when running pulumi update. Not sure how I missed that originally.

### Testing
- make build and install
- Tested locally using Dotnet SDK

### DIFF
--- a/provider/pkg/provider/drift_schedules.go
+++ b/provider/pkg/provider/drift_schedules.go
@@ -63,7 +63,21 @@ func ToPulumiServiceDriftScheduleInput(properties *structpb.Struct) (*PulumiServ
 }
 
 func (st *PulumiServiceDriftScheduleResource) Diff(req *pulumirpc.DiffRequest) (*pulumirpc.DiffResponse, error) {
-	return ScheduleSharedDiff(req)
+	olds, err := plugin.UnmarshalProperties(req.GetOlds(), plugin.MarshalOptions{KeepUnknowns: false, SkipNulls: true})
+	if err != nil {
+		return nil, err
+	}
+
+	news, err := plugin.UnmarshalProperties(req.GetNews(), plugin.MarshalOptions{KeepUnknowns: true, SkipNulls: true})
+	if err != nil {
+		return nil, err
+	}
+
+	if !news["autoRemediate"].HasValue() {
+		news["autoRemediate"] = resource.NewBoolProperty(false)
+	}
+
+	return ScheduleSharedDiffMaps(olds, news)
 }
 
 func (st *PulumiServiceDriftScheduleResource) Delete(req *pulumirpc.DeleteRequest) (*pbempty.Empty, error) {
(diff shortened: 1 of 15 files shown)


IaroslavTitov:
PR checks are failing, so I made https://github.com/pulumi/pulumi-pulumiservice/pull/261 to fix the wf

provider/pkg/provider/drift_schedules.go
```diff
@@ -32,7 +32,9 @@ func (i *PulumiServiceDriftScheduleInput) ToPropertyMap() resource.PropertyMap {
 	propertyMap := StackToPropertyMap(i.Stack)
 
 	propertyMap["scheduleCron"] = resource.NewPropertyValue(i.ScheduleCron)
-	propertyMap["autoRemediate"] = resource.NewPropertyValue(i.AutoRemediate)
+	if i.AutoRemediate != nil {
+		propertyMap["autoRemediate"] = resource.NewPropertyValue(*i.AutoRemediate)
+	}
```
komalali, line 37: You should be able to avoid doing this by setting the default in the [schema](https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/cmd/pulumi-resource-pulumiservice/schema.json#L1110-L1113) like some [properties in deploymentSettings](https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/cmd/pulumi-resource-pulumiservice/schema.json#L303)

IaroslavTitov: Good point, didn't realize default values existed in schema. Updated the PR to use default values, but sadly then had to adjust Diff logic, as it turns out in Diff's GetNews() doesn't provide default values, so second and beyond pulumi up would always show diffs

komalali: Huh, I find that surprising. We use a similar pattern in deploymentSettings and don't have to do this in the diff function. Looking at the generated SDKs it would be very surprising if this didn't work since they all provide the defaults if they're not specified. Which SDK were you testing with?

komalali: Wondering if maybe the SDK wasn't regen'd before you tested.

komalali (approved):
Curious about the question I asked about News not including the default values, but doesn't hurt to be extra specific anyway!

provider/pkg/provider/drift_schedules.go
IaroslavTitov: I used Dotnet SDK
I ran pulumi up on schedules without the booleans, and they were filled with false successfully (indicating that the schema change worked), but when I ran pulumi up again it showed the diff.
I will try to clean SDK and regenerate to make sure before I merge, just in case

IaroslavTitov: Confirmed by cleaning SDK, building this branch -> no change in diff, then removed explicit setting to false -> diff shows up. Sounds kind of like a bug

Issue #225, opened 2024-02-09
Add ESC Environment Resource to the Pulumi Cloud Provider


desteves:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

I have a Pulumi program with AWS Resources; in this same program, I'd like to create a new ESC Environment. I can do so using axios and [this REST API](https://www.pulumi.com/docs/pulumi-cloud/cloud-rest-api/#create-environment), but it's messy.

### Affected area/feature

Pulumi Service

lukehoban:
Related to https://github.com/pulumi/pulumi-pulumiservice/issues/179 and https://github.com/pulumi/pulumi-pulumiservice/issues/63.

PR #271, opened 2024-05-10
Adding Environment Resource

IaroslavTitov:
### Summary

- Adding Environment resource
- Adding dependency on ESC client
- Fixes [255](https://github.com/pulumi/pulumi-pulumiservice/issues/225)

### Testing

Tested in dev stack using below Pulumi Program (Dotnet SDK), confirmed that things like imports, opening environment work as expected.

```
    String yaml = """
    imports:
      - dev-stacks
    values:
      aws:
        secrets:
          fn::open::aws-secrets:
            region: us-west-2
            login: ${aws.creds}
            get:
              allSecrets:
                secretId: iaro-dev-stack/pulumi-service
      secrets:
        fn::fromJSON: ${aws.secrets.allSecrets}
      pulumiConfig:
        gitHubOAuthID: ${secrets.gitHubOAuthID}
        gitHubOAuthSecret: ${secrets.gitHubOAuthSecret}
    """;

    var environ = new Environment(
        "Iaro's environment",
        new EnvironmentArgs {
            Name = "IaroEnv",
            Yaml = new StringAsset(yaml)
        }
    );
```


provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1219,6 +1219,47 @@
...
+          "description": "Organization name.",
+          "type": "string"
+        },
+        "name": {
+          "description": "Environment name.",
+          "type": "string"
+        },
+        "yaml": {
+          "description": "Environment's yaml file.",
+          "$ref": "pulumi.json#/Asset"
```
pgavlin, line 1236: Should make absolutely certain that assets work with output properties. Not 100% sure about that.

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1219,6 +1219,47 @@
...
+        }
+      },
+      "required": [
+        "organization",
+        "name",
```
pgavlin, line 1241: In other providers we have a pattern of allowing the name to be optional and pulling it from the URN during `Check` if it's unspecified. I think it's okay to leave it required here--I think we can make it optional + compute it in Check in the future w/o a breaking change if we choose.

provider/pkg/provider/environment.go
```diff
@@ -0,0 +1,271 @@
...
+	if diagnostics != nil {
+		return nil, fmt.Errorf("failed to create environment, yaml code failed following checks: %+v", diagnostics)
+	}
+
+	// Then create environment, and update it with yaml provided. ESC API architecture doesn't let you do it in one call
+	err = st.client.CreateEnvironment(context.Background(), input.OrgName, input.EnvName)
+	if err != nil {
+		return nil, err
+	}
+	diagnostics, err = st.client.UpdateEnvironment(context.Background(), input.OrgName, input.EnvName, input.Yaml, ETAG)
```
pgavlin, line 153: Should be able to just pass the empty string instead of `ETAG` here.

provider/pkg/provider/environment.go
```diff
@@ -0,0 +1,271 @@
...
+	return &pulumirpc.CheckResponse{Inputs: req.GetNews(), Failures: failures}, nil
+}
+
+func (st *PulumiServiceEnvironmentResource) Update(req *pulumirpc.UpdateRequest) (*pulumirpc.UpdateResponse, error) {
+	input, err := ToPulumiServiceEnvironmentInput(req.GetNews())
+	if err != nil {
+		return nil, err
+	}
+
+	diagnostics, err := st.client.UpdateEnvironment(context.Background(), input.OrgName, input.EnvName, input.Yaml, ETAG)
```
pgavlin, line 205: Might want to store the actual ETag in the state and pass it in here. That would be something of a double-edged sword, as it would ensure that manually-edited environments need to be refreshed prior to an update.

Otherwise this should just pass the empty string instead of `ETAG`.

provider/pkg/provider/provider.go
```diff
@@ -120,6 +123,8 @@ func (k *pulumiserviceProvider) Configure(_ context.Context, req *pulumirpc.Conf
 	}
 	client, err := pulumiapi.NewClient(&httpClient, *token, *url)
 
+	escClient := esc_client.New(fmt.Sprintf("esc-sdk/1 (%s; %s)", version.Version, runtime.GOOS), *url, *token, false)
```
pgavlin, line 126: Probably want to use a distinct user agent here so that we can distinguish provider traffic from other traffic

```suggestion
	escClient := esc_client.New(fmt.Sprintf("esc-sdk/1 (%s; %s)", version.Version, runtime.GOOS), *url, *token, false)
```

provider/cmd/pulumi-resource-pulumiservice/schema.json
IaroslavTitov: Confirmed it works with output using Random and below snippet of code:

```
var random = new RandomString("rand", new RandomStringArgs
    {
        Length = 5,
        Special = false,
    });

    Output<AssetOrArchive> asset = random.Result.Apply(res => {
        String yaml = """
            imports:
              - dev-stacks
            values:
              aws:
                secrets:
                  fn::open::aws-secrets:
                    region: us-west-2
                    login: ${aws.creds}
                    get:
                      allSecrets:
                        secretId: iaro-dev-stack/pulumi-service
              secrets:
                fn::fromJSON: ${aws.secrets.allSecrets}
              pulumiConfig:
                gitHubOAuthID: ${secrets.gitHubOAuthID}
                gitHubOAuthSecret: ${secrets.gitHubOAuthSecret}
                randomString: 
        """ + res;
        AssetOrArchive asset = new StringAsset(yaml);
        return asset;
        }
    );

    var environ = new Pulumi.PulumiService.Environment(
        "Iaro's environment",
        new EnvironmentArgs {
            Name = "IaroEnv",
            Yaml = asset
        }
    );
```

When I go to see the environment in console it has a random blurb `zbhrh` and the same in Random's resource properties.

provider/cmd/pulumi-resource-pulumiservice/schema.json
IaroslavTitov: Do I understand it right that the name is not stored separately as a property and just pulled from id every time? I'm not sure I understand the benefit apart from saving a bit of DB space... Would you link an example please?

provider/pkg/provider/environment.go
IaroslavTitov: Replaced with empty string, I re-read ESC client comments and realized I misunderstood the point of ETAG, I thought it was like a source of change, to prevent simultaneous updates.

provider/pkg/provider/environment.go
IaroslavTitov: Same as above, replaced with empty string

provider/pkg/provider/provider.go
IaroslavTitov: Replaced with `"pulumicloud-provider/1 (%s; %s)"`

provider/cmd/pulumi-resource-pulumiservice/schema.json
pgavlin: Ah, I should clarify--I meant that I'm not sure that the output property `Environment.Yaml` will work with assets. I'd try passing the `Yaml` output of one environment to the `Yaml` input of another.

provider/cmd/pulumi-resource-pulumiservice/schema.json
pgavlin: It's a UX benefit--it keeps users from repeating names and allows the provider to add entropy in order to avoid collisions between multiple deployments of the same Pulumi program.

For example, the following code will create a bucket named `foo-<some random suffix>`:

```typescript
const b = new aws.s3.Bucket("foo");
```

But this code will create a bucket named `foo`:

```typescript
const b = new aws.s3.Bucket("foo", { bucket: "foo" });
```

In the first example, the provider's `Check` method will pick `foo` out of the "name" portion of the resource's URN, attach some entropy to the end, and inject the final name into the result (unless there is a name present in the state, in which case it will use that name).

I'm not sure that this is behavior we _want_ for the `Environment` resource, but it's something that's worth considering.

IaroslavTitov: Ack, makes sense, thank you for the laid out example

I think it might look ugly if we by default have `EnvironmentName-abc123` in the console after customer named the resource `EnvironmentName`. For things like s3 buckets you probably don't care what the name is, for environment I think people would care more. But also if customers are creating environments via provider, they aren't likely to go look at it in the console, so not sure.

Let me know if you'd prefer I add this logic, I don't know enough about providers yet to have a strong opinion

provider/cmd/pulumi-resource-pulumiservice/schema.json
IaroslavTitov: Oooh, gotcha, didn't understand properly
Confirmed that this works:
```
var environ = new Pulumi.PulumiService.Environment(
        "Iaro's environment",
        new EnvironmentArgs {
            Name = "IaroEnv",
            Yaml = asset
        }
    );

    var environ2 = new Pulumi.PulumiService.Environment(
        "Iaro's environment 2",
        new EnvironmentArgs {
            Name = "IaroEnv2",
            Yaml = environ.Yaml.Apply(yaml => yaml)
        }
    );
```

komalali: FWIW @IaroslavTitov you shouldn't have to do this:
```
Yaml = environ.Yaml.Apply(yaml => yaml)
```

It can just be:
```
Yaml = environ.Yaml
```

provider/cmd/pulumi-resource-pulumiservice/schema.json
komalali: FWIW @pgavlin we don't have autonaming for any resources in this provider yet. I think let's leave it as-is and address auto-naming wholesale?

provider/pkg/provider/provider.go
komalali: @IaroslavTitov let's keep it the actual name `provider-pulumiservice` or just `provider`.

provider/cmd/pulumi-resource-pulumiservice/schema.json
pgavlin: SGTM

provider/cmd/pulumi-resource-pulumiservice/schema.json
IaroslavTitov: Oh, makes sense, got stuck in the Apply mindset

provider/pkg/provider/provider.go
IaroslavTitov: Sounds good, updated

pgavlin:
High-level question:

If my environment definition contains secrets, are those stored in plaintext unless the entire environment definition is marked secret? For example, in the following program, will `foo`'s value be stored in plaintext?

```typescript
const env = new Environment("env", { yaml: `{ values: { foo: { fn::secret: hunter2 } } }` });
```

IaroslavTitov:
Tried this, shows up as plaintext in the Resources tab. (In Environments, shows up as ciphertext)

Unfortunately, I don't think we can avoid storing it as plaintext, in order for import to work, this is the same issue as with Deployment Settings. We can add some frontend logic to *** them out maybe? Still obv wouldn't actually hide the secrets, just visually.

pgavlin:
So users can manually mark the environment def as secret in their Pulumi program:

```typescript
const env = new Environment("env", { yaml: pulumi.secret(`{ values: { foo: { fn::secret: hunter2 } } }`) });
```

That will encrypt the YAML in the statefile. We might also be able to do this automatically as part of `Check`, but I'm not certain. `Check` might be able to wrap the input YAML in a `Secret` prior to returning the inputs to the engine. I think that's worth a look.

> Unfortunately, I don't think we can avoid storing it as plaintext, in order for import to work, this is the same issue as with Deployment Settings

Yeah we need to chat more about our approach to diffing with secrets. I think that we have several options here, but all of them will involve tradeoffs.

IaroslavTitov:
> `Check` might be able to wrap the input YAML in a `Secret` prior

Ooh, I did not think of that, I think that shouldn't be hard, I'll look into it and update the PR.

IaroslavTitov:
Updated to force yaml to always be a secret property. 

I then went down the rabbit hole trying to solve noisy diff when user has extra spaces in the yaml, which ESC then removes and causes diff. Sadly, found no way of eliminating this issue with secrets, as Diff receives only hash of a secret Asset, not full text and this can only be compared exactly, and not partially. It is a rare chance anyone will hit that issue though (I totally accidentally stumbled into it), so I think it's fine.

komalali:
@IaroslavTitov can you resolve the conflicts please?

IaroslavTitov:
> @IaroslavTitov can you resolve the conflicts please?

Done!

komalali (approved):

PR #272, opened 2024-05-10
Added examples on how to use schedules

IaroslavTitov:
### Summary
- Added C# example on creating schedules
- Added Typescript example on creating schedules
- Added Yaml example on creating schedules
- Added them to list of tests
- Fixes [274](https://github.com/pulumi/pulumi-pulumiservice/issues/274)

### Testing
- Tested all in local projects


komalali:
@IaroslavTitov I just made a change to the service that now validates if the stack has deployment settings and returns an error if it doesn't. So you may need to update these tests to also include deployment setting resources.

examples/cs-schedules/MyStack.cs
```diff
@@ -0,0 +1,92 @@
...
+        );
+
+        // Schedule that runs drift every Sunday midnight, but does NOT remediate it
+        var drift = new DriftSchedule(
+            "drift-schedule",
+            new DriftScheduleArgs
+            {
+                Organization = "service-provider-test-org",
+                Project = "cs-schedules",
+                Stack = "test-stack",
```
komalali, line 34: You can use the output from deployment settings as an input here to create the dependency chain to ensure that the schedules are always created *after* the settings.
```suggestion
                Organization = settings.Organization,
                Project = settings.Project,
                Stack = settings.Stack,
```

examples/cs-schedules/MyStack.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+        );
+
+        // Schedule to destroy stack resources on Jan 1, 2026, but NOT delete the stack itself
+        var ttl = new TtlSchedule(
+            "ttl-schedule",
+            new TtlScheduleArgs{
+                Organization = "service-provider-test-org",
+                Project = "cs-schedules",
+                Stack = "test-stack",
```
komalali, line 50: Same idea everywhere below

```suggestion
                Organization = settings.Organization,
                Project = settings.Project,
                Stack = settings.Stack,
```

examples/cs-schedules/MyStack.cs
IaroslavTitov: Yeah, I used dependsOn below to make dependency explicit, and mentioned it above in comment to explain to the reader.

I can do the settings properties too, agree that it's cleaner. However, in this case, since it's an example I think explicitly showing dependency is better, but can change, don't mind either way

komalali: Ah whoops I see it now. Fine to leave as-is.

komalali (approved):

PR #289, opened 2024-05-28
Adding revision to Environment resource

IaroslavTitov:
### Summary
- Adding a revision property to Environment resource
- Moving to a newer esc client added more dummy methods into the test file

### Testing
- Tested in devstack using Dotnet API
- revision number changes as expected

### DIFF
--- a/provider/pkg/provider/environment_test.go
+++ b/provider/pkg/provider/environment_test.go
@@ -14,14 +14,26 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-type getEnvironmentFunc func(ctx context.Context, orgName string, envName string, decrypt bool) (yaml []byte, etag string, err error)
+type getEnvironmentFunc func(ctx context.Context, orgName string, envName string, version string, decrypt bool) (yaml []byte, etag string, revision int, err error)
 
 type EscClientMock struct {
 	getEnvironmentFunc getEnvironmentFunc
 }
 
-func (c *EscClientMock) GetEnvironment(ctx context.Context, orgName string, envName string, decrypt bool) (yaml []byte, etag string, err error) {
-	return c.getEnvironmentFunc(ctx, orgName, envName, decrypt)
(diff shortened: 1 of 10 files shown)


provider/pkg/provider/environment_test.go
```diff
@@ -14,14 +14,26 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-type getEnvironmentFunc func(ctx context.Context, orgName string, envName string, decrypt bool) (yaml []byte, etag string, err error)
+type getEnvironmentFunc func(ctx context.Context, orgName string, envName string, version string, decrypt bool) (yaml []byte, etag string, revision int, err error)
```
komalali, line 17: What is the `version` here?

sdk/nodejs/environment.ts
```diff
@@ -70,9 +74,11 @@ export class Environment extends pulumi.CustomResource {
             resourceInputs["name"] = args ? args.name : undefined;
             resourceInputs["organization"] = args ? args.organization : undefined;
             resourceInputs["yaml"] = args ? args.yaml : undefined;
+            resourceInputs["revision"] = undefined /*out*/;
```
komalali, line 77: This is odd... why is it in the inputs section?

provider/pkg/provider/environment_test.go
IaroslavTitov: This function has to match the signature of GetEnvironment method inside ESC client. I bumped the version and signature of the method changed, so had to change to match as well.

sdk/nodejs/environment.ts
komalali: I think this is fine, I'm just not used to seeing outputs only.

komalali (approved):

sdk/nodejs/environment.ts
IaroslavTitov: I'm guessing because it's an output only property?

PR #292, opened 2024-05-30
Adding EnvironmentVersionTag resource

IaroslavTitov:
### Summary
- Adding EnvironmentVersionTag resource

### Testing
- Tested locally using Dotnet SDK

```
    var environ = new Pulumi.PulumiService.Environment(
        "Iaro's environment",
        new EnvironmentArgs {
            Organization = "service-provider-test-org",
            Name = "IaroEnv",
            Yaml = new StringAsset(yaml)
        }
    );

    var tag = new Pulumi.PulumiService.EnvironmentVersionTag(
        "Iaro's version tag",
        new EnvironmentVersionTagArgs {
            Organization = "service-provider-test-org",
            Environment = "IaroEnv",
            TagName = "SuperTag",
            Revision = environ.Revision
        }
    );
```


provider/pkg/provider/environment_version_tags.go
```diff
@@ -0,0 +1,162 @@
...
+	}, nil
+}
+
+func (evt *PulumiServiceEnvironmentVersionTagResource) Check(req *pulumirpc.CheckRequest) (*pulumirpc.CheckResponse, error) {
+	return &pulumirpc.CheckResponse{Inputs: req.News, Failures: nil}, nil
+}
+
+func (evt *PulumiServiceEnvironmentVersionTagResource) Update(req *pulumirpc.UpdateRequest) (*pulumirpc.UpdateResponse, error) {
+	// all updates are destructive, so we just call Create.
+	return nil, fmt.Errorf("unexpected call to update, expected create to be called instead")
```
pgavlin, line 118: We definitely don't want tags to always be replaced. Every change to a tag besides its name should be an update.

pgavlin (changes requested):

provider/pkg/provider/environment_version_tags.go
IaroslavTitov: I followed the way stack tags are implemented [here](https://github.com/pulumi/pulumi-pulumiservice/blob/f13da7e94e5f83093da265d3a3c7c4d842375ba9/provider/pkg/provider/stack_tags.go#L133)

I'll rework it to have updates

pgavlin (approved):

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1344,6 +1344,33 @@
         "environment",
         "permission"
       ]
+    },
+    "pulumiservice:index:EnvironmentVersionTag": {
```
komalali, line 1348: Does the resource not have any outputs?

komalali: We should mirror all inputs as outputs, at least

IaroslavTitov: Good catch, somehow works without it lol
I'll add real quick

komalali (approved):

PR #293, opened 2024-06-03
Adding examples on using revision tags for environments

IaroslavTitov:
### Summary
- Adding examples on using revision tags for environments

### Testing
- All the examples automatically run as tests


examples/py-environments/__main__.py
```diff
@@ -15,3 +16,26 @@
...
+	organization=environment.organization,
+	environment=environment.name,
+	tag_name=environment.revision.apply(
+    lambda revision: "v" + str(revision)
+  ),
+	revision=environment.revision,
+  opts=ResourceOptions(
+    retain_on_delete=True
+  )
+)
```
komalali, line 41: nit: newline
```suggestion
)

```

examples/ts-environments/index.ts
```diff
@@ -12,3 +12,21 @@ var environment = new service.Environment("testing-environment", {
...
+
+// A tag that will be placed on each new version, and remain on old revisions
+var versionTag = new service.EnvironmentVersionTag("VersionTag", {
+  organization: environment.organization,
+  environment: environment.name,
+  tagName: environment.revision.apply((rev: number) => "v"+rev),
+  revision: environment.revision
+}, {
+  retainOnDelete: true
+})
```
komalali, line 32:
```suggestion
})

```

examples/yaml-environments/Pulumi.yaml
```diff
@@ -8,9 +8,25 @@ resources:
...
+      revision: ${testing-environment.revision}
+  versionTag:
+    type: pulumiservice:EnvironmentVersionTag
+    properties:
+      organization: ${testing-environment.organization}
+      environment: ${testing-environment.name}
+      tagName: v${testing-environment.revision}
+      revision: ${testing-environment.revision}
+    options:
+      retainOnDelete: true
```
komalali, line 32:
```suggestion
      retainOnDelete: true

```

examples/ts-environments/index.ts
komalali, line 32: You had mentioned yesterday that this resource worked without the outputs... but we're not really using outputs in any of these examples so they're not really tested. Something like `export const tag = versionTag.tagName;` would test it.

komalali (commented):
These haven't been added to the test files... was that intentional?

Ah I just realized these are additions to existing examples

IaroslavTitov:
> These haven't been added to the test files... was that intentional?

I didn't really add new tests, just extended existing ones, so there's nothing to add really

komalali (approved):

examples/ts-environments/index.ts
IaroslavTitov: Yeah, I realized that I never actually used any properties. I tested and checked that after adding to the schema the properties have appeared. (It's easy to check in Dotnet)

PR #298, opened 2024-06-05
Adding update functionality to Deployment Settings resource

IaroslavTitov:
### Summary
- Adding update functionaity using the new replace endpoint for DeploymentSettings

### Testing
- Manual test using Dotnet SDK

### DIFF
--- a/provider/pkg/provider/deployment_settings.go
+++ b/provider/pkg/provider/deployment_settings.go
@@ -488,7 +488,50 @@ func getSecretOrStringValue(prop resource.PropertyValue) string {
 }
 
 func (ds *PulumiServiceDeploymentSettingsResource) Diff(req *pulumirpc.DiffRequest) (*pulumirpc.DiffResponse, error) {
-	return considerAllChangesReplaces(req)
+	olds, err := plugin.UnmarshalProperties(req.GetOlds(), plugin.MarshalOptions{KeepUnknowns: true, SkipNulls: true})
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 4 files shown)


provider/pkg/provider/deployment_settings.go
```diff
@@ -488,7 +488,51 @@ func getSecretOrStringValue(prop resource.PropertyValue) string {
...
+
+	dd := plugin.NewDetailedDiffFromObjectDiff(diffs, false)
+
+	detailedDiffs := map[string]*pulumirpc.PropertyDiff{}
+	replaces := []string(nil)
+	replaceProperties := map[string]bool{
+		"organization":  true,
+		"project":       true,
+		"stack":         true,
+		"sourceContext": true,
```
komalali, line 516: I don't think we need to replace on source. We should still be able to just update.

provider/pkg/provider/deployment_settings.go
```diff
@@ -488,7 +488,51 @@ func getSecretOrStringValue(prop resource.PropertyValue) string {
...
+			InputDiff: v.InputDiff,
+		}
+	}
+
+	return &pulumirpc.DiffResponse{
+		Changes:             pulumirpc.DiffResponse_DIFF_SOME,
+		Replaces:            replaces,
+		DetailedDiff:        detailedDiffs,
+		HasDetailedDiff:     true,
+		DeleteBeforeReplace: len(replaces) > 0,
```
komalali, line 534: FWIW you can just set this to `true`, it'll only be needed in the case of a replace and we'll want it to be true if it is.

komalali (approved):

PR #308, opened 2024-06-06
Updating deprecated goreleaser flag in main workflow

IaroslavTitov:
Another goreleaser deprecation, source - https://goreleaser.com/deprecations/?h=skip+validate#-skip

This is needed to fix main workflow

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -197,7 +197,7 @@ jobs:
     - name: Run GoReleaser
       uses: goreleaser/goreleaser-action@v2
       with:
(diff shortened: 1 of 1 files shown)


komalali (approved):

komalali:
Did we happen to check if there are any other workflows using goreleaser?

IaroslavTitov:
> Did we happen to check if there are any other workflows using goreleaser?

Just main and release are, so this will hopefully be the last goreleaser change. I was rushing to get the release out yesterday, so didn't really even look at main

PR #309, opened 2024-06-06
Moving from GetOlds to GetOldInputs to avoid having to manually remove properties

IaroslavTitov:
### Summary
- Moving all Diff methods to compare inputs rather than final properties
- This way we don't need to remove properties to avoid a diff, code is more straightforward

### Testing
- All the resources are checked for diff by integ tests
- Testing envrionment resource manually


komalali (approved):
@IaroslavTitov why did you have to check the environment resource manually?

IaroslavTitov:
> @IaroslavTitov why did you have to check the environment resource manually?

Oh, I didn't necessarily need to, I just had it locally already, so decided to test on it

PR #310, opened 2024-06-07
Fixing Schedules import

IaroslavTitov:
### Summary
- Reworked Read method to read using values from id, which allows import of values

### Testing
- Tested using below commands:
- `pulumi import pulumiservice:index:DriftSchedule importedthing2 IaroslavTitov/PulumiDotnet/SdkTest5/ttl/c940d27b-d563-4f07-8833-9f2d14baa32b`
- `pulumi import pulumiservice:index:TtlSchedule importedthing2 IaroslavTitov/PulumiDotnet/SdkTest5/ttl/c940d27b-d563-4f07-8833-9f2d14baa32b`
- `pulumi import pulumiservice:index:DeploymentSchedule importedthing3 IaroslavTitov/PulumiDotnet/SdkTest5/539430ce-96e9-4228-a5a2-5a3cd138d0df`

I had to grab schedule id's using browser tools, but we already have an issue to address that -  https://github.com/pulumi/pulumi-service/issues/20094

### DIFF
--- a/provider/pkg/provider/deployment_schedules.go
+++ b/provider/pkg/provider/deployment_schedules.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/pulumi/pulumi-pulumiservice/provider/pkg/internal/pulumiapi"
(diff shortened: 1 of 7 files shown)


provider/pkg/provider/deployment_schedules.go
```diff
@@ -322,24 +323,35 @@ func (st *PulumiServiceDeploymentScheduleResource) Update(req *pulumirpc.UpdateR
...
 
-	scheduleID, err := st.client.GetSchedule(context.Background(), output.Stack, output.ScheduleID)
+	scheduleResponse, err := st.client.GetSchedule(context.Background(), *stack, *scheduleID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read DeploymentSchedule (%q): %w", req.Id, err)
 	}
-	if scheduleID == nil {
-		// if the tag doesn't exist, then return empty response
+	if scheduleResponse == nil {
+		// if schedule doesn't exist, then return empty response
```
komalali, line 336: Might be worth clarifying this comment to state that this deletes the resource from state.
```suggestion
		// if schedule doesn't exist, then return empty response to delete it from state
```

komalali (approved):

provider/pkg/provider/deployment_schedules.go
IaroslavTitov: Good point, will update all 3 places

PR #320, opened 2024-06-14
Fixing Deployment Secrets - Parallel Values

IaroslavTitov:
Note: this is an alternative approach for PR - https://github.com/pulumi/pulumi-pulumiservice/pull/325

### Summary

- Added logic to save cipher and plaintext separately, allowing comparison on refresh of just ciphertext, fixing https://github.com/pulumi/pulumi-pulumiservice/issues/123
- Import now works as well, including code generation (with dummy values for secrets)
- Migrated to new PUT API and updated client to actually return DeploymentSettings

### Testing
- Tested pulumi up, refresh, import and up from previous version of the provider (for unchanged DS inputs, migrating to this new way of saving will require refresh and then up)

Example TS program (Sadly can't use Dotnet, due to bug with maps):
```
const settings = new service.DeploymentSettings("deployment_settings", {
  organization: "IaroslavTitov",
  project: "PulumiDotnet",
  stack: "SdkTest5",
  operationContext: {
    environmentVariables: {
      TEST_VAR: "fooa",
      SECRET_VAR: config.requireSecret("my_secret"),
    }
  },
  sourceContext: {
    git: {
        repoUrl: "https://github.com/pulumi/deploy-demos.git",
        branch: "refs/heads/main",
        repoDir: "pulumi-programs/simple-resource",
        gitAuth: {
            sshAuth: {
                sshPrivateKey: "privatekey",
                password: secret,
            }
        }
    }
}
});

Secret resource values end up split into cipher and plaintext like this:
```
        "sshPrivateKeyCipher": "AAABAJrXK3j7+jgXLJI9Q30JnAFDxek6yUyYudh3h2SU4/dQKyBl",
        "sshPrivateKey": {
          "ciphertext": "AAABAKlREY4tsjd1bbWc8JXPaABRqG6BeVxfUkhO1ob70Q=="
        }
```
twin values for each secret. Environment variables ciphers are stored in a separate variable `environmentVariablesCipher`

Passwords and sshKey are forced into twin secrets, Environment Variables are optionally twin secrets, everything else uses normal Pulumi workdlows, because they are not secret in Pulumi Service.

Import of the above code generates successfully with dummy values for secrets:
```
const ds1 = new pulumiservice.DeploymentSettings("ds1", {
    operationContext: {
        environmentVariables: {
            SECRET_VAR: pulumi.secret("<REPLACE WITH ACTUAL SECRET VALUE>"),
            TEST_VAR: "fooa",
        },
    },
    organization: "IaroslavTitov",
    project: "PulumiDotnet",
    sourceContext: {
        git: {
            branch: "refs/heads/main",
            gitAuth: {
                sshAuth: {
                    password: pulumi.secret("<REPLACE WITH ACTUAL SECRET VALUE>"),
                    sshPrivateKey: pulumi.secret("<REPLACE WITH ACTUAL SECRET VALUE>"),
                },
            },
            repoDir: "pulumi-programs/simple-resource",
            repoUrl: "https://github.com/pulumi/deploy-demos.git",
        },
    },
    stack: "SdkTest5",
}, {
    protect: true,
});
```

### DIFF
--- a/provider/pkg/internal/pulumiapi/deployment_settings.go
+++ b/provider/pkg/internal/pulumiapi/deployment_settings.go
@@ -2,6 +2,7 @@ package pulumiapi
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"path"
@@ -10,8 +11,8 @@ import (
 )
 
 type DeploymentSettingsClient interface {
-	CreateDeploymentSettings(ctx context.Context, stack StackName, ds DeploymentSettings) error
-	UpdateDeploymentSettings(ctx context.Context, stack StackName, ds DeploymentSettings) error
+	CreateDeploymentSettings(ctx context.Context, stack StackName, ds DeploymentSettings) (*DeploymentSettings, error)
+	UpdateDeploymentSettings(ctx context.Context, stack StackName, ds DeploymentSettings) (*DeploymentSettings, error)
 	GetDeploymentSettings(ctx context.Context, stack StackName) (*DeploymentSettings, error)
 	DeleteDeploymentSettings(ctx context.Context, stack StackName) error
 }
@@ -19,16 +20,17 @@ type DeploymentSettingsClient interface {
 type DeploymentSettings struct {
 	OperationContext *OperationContext        `json:"operationContext,omitempty"`
 	GitHub           *GitHubConfiguration     `json:"gitHub,omitempty"`
-	SourceContext    *apitype.SourceContext   `json:"sourceContext,omitempty"`
+	SourceContext    *SourceContext           `json:"sourceContext,omitempty"`
 	ExecutorContext  *apitype.ExecutorContext `json:"executorContext,omitempty"`
 	AgentPoolId      string                   `json:"agentPoolId,omitempty"`
+	Source           *string                  `json:"source,omitempty"`
 }
 
 type OperationContext struct {
-	Options              *OperationContextOptions       `json:"options,omitempty"`
-	PreRunCommands       []string                       `json:"PreRunCommands,omitempty"`
-	EnvironmentVariables map[string]apitype.SecretValue `json:"environmentVariables,omitempty"`
-	OIDC                 *OIDCConfiguration             `json:"oidc,omitempty"`
+	Options              *OperationContextOptions `json:"options,omitempty"`
+	PreRunCommands       []string                 `json:"PreRunCommands,omitempty"`
+	EnvironmentVariables map[string]SecretValue   `json:"environmentVariables,omitempty"`
+	OIDC                 *OIDCConfiguration       `json:"oidc,omitempty"`
 }
 
 type OIDCConfiguration struct {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: provider/cmd/pulumi-resource-pulumiservice/schema.json [hunk 1]
```diff
@@ -323,9 +323,9 @@
           "description": "Shell commands to run before the Pulumi operation executes."
         },
         "environmentVariables": {
-          "type": "object",
-          "additionalProperties": {
-            "type": "string"
+          "type": "array",
+          "items": {
+            "$ref": "#/types/pulumiservice:index:EnvironmentVariable"
```
komalali, line 328: This is a breaking change, and we should generally avoid making breaking changes to the highest extent possible. Is the reason for the type change only to avoid https://github.com/pulumi/pulumi-dotnet/issues/22?

IaroslavTitov: Yeah, breaking change is not great, I assumed building a backward compatible way would be overkill (and take like a week)
Two reasons - one is to avoid the bug, another is to allow twin values. Right now I'm keeping key plaintext, while value is stored using twin value: `valuePlain` and 'valueCipher'. Because my code knows the string `value`, I can add Plain or Cipher to it and find the property. However, if I had a map like before, I'd need two key value pairs, instead of two properties, and then I'd have to append keys with Plain or Cipher. And if I do that, I have no way of telling what the original key actually was ':D I could parse all and guess, but then there's always a chance someone calls their key "keyPlain" and that would screw it up further ':D

provider/pkg/internal/pulumiapi/deployment_settings.go
```diff
@@ -74,22 +76,87 @@ type GitHubConfiguration struct {
...
+type BasicAuth struct {
+	UserName SecretValue `json:"userName"`
+	Password SecretValue `json:"password"`
+}
+
+type SecretValue struct {
+	Value  string // Plaintext if Secret is false; ciphertext otherwise.
+	Secret bool
+}
+
```
komalali, line 111: Doesn't this return both `Value` and `Ciphertext`?

provider/cmd/pulumi-resource-pulumiservice/schema.json
komalali: I'm not sure I understand why this implementation detail needs to be surfaced to the user. Or why the map couldn't be a map of string to objects in the implementation rather than an array of objects like it is in the service.

I don't think a breaking change is acceptable here, let's chat and we can discuss the issues you're running into and what other strategies we have to get around it.

komalali (changes requested):

provider/pkg/internal/pulumiapi/deployment_settings.go
IaroslavTitov: No, Pulumi Service only returns either plaintext or ciphertext, with the Secret bool indicating which one. But since we opted not to have the ability to return plaintext on these secret deployment settings, it only ever returns ciphertext.

This whole struct tree is taken from sdk, I unfrotunately had to copy it in to override the marshaling/unmarshaling logic

IaroslavTitov:
Updated the code.
This way it's not a breaking change, which is well worth the extra bit of complexity, I just didn't see this way until today.

The env vars are now stored as map of string to object, inside which ciphertext and plaintext are packed. Import now needs a separate flag for serialization into Inputs, separate from serialization into Properties, but it's a pretty small change.

IaroslavTitov:
Ran into an issue with python tests, filed https://github.com/pulumi/pulumi-pulumiservice/issues/322 to address later, looks like another bug with providers

sdk/go/pulumiservice/environment.go
```diff
@@ -0,0 +1,250 @@
+// Code generated by pulumi-language-go DO NOT EDIT.
```
komalali, line 1: This diff is surprising in this PR. Did we forget to include this earlier?

IaroslavTitov: Yep, the go files only didn't generate for Environment resources, Stoo pinged me about it. Looking into it, a while ago I updated .gitignore and it's covering too much, which caused this. I'll have to address the gitignore issue, but figured should include the missed files asap

komalali: Can we pull this out into a separate PR and get that merged and released? I'd rather not hold that up on this PR.

komalali:
Can we close this out @IaroslavTitov ?

IaroslavTitov:
Closing as this was just an alternate version

PR #325, opened 2024-06-20
Fixing Deployment Secrets - Plaintext in Inputs

IaroslavTitov:
### Summary

- Added logic to save ciphertext into the output properties for secret values, allowing comparison on refresh of just ciphertext, fixing https://github.com/pulumi/pulumi-pulumiservice/issues/123
- Import now works as well, including code generation (with dummy values for secrets)
- Migrated to new PUT API and updated client to actually return DeploymentSettings

### Testing
- Tested pulumi up, refresh, import and up from previous version of the provider (for unchanged DS inputs, migrating to this new way of saving will require refresh and then up)

Example TS program (Sadly can't use Dotnet, due to bug with maps):
```
const settings = new service.DeploymentSettings("deployment_settings", {
  organization: "IaroslavTitov",
  project: "PulumiDotnet",
  stack: "SdkTest5",
  operationContext: {
    environmentVariables: {
      TEST_VAR: "fooa",
      SECRET_VAR: config.requireSecret("my_secret"),
    }
  },
  sourceContext: {
    git: {
        repoUrl: "https://github.com/pulumi/deploy-demos.git",
        branch: "refs/heads/main",
        repoDir: "pulumi-programs/simple-resource",
        gitAuth: {
            sshAuth: {
                sshPrivateKey: "privatekey",
                password: secret,
            }
        }
    }
}
});
```

Secret resource values end up with just cipher, while plaintext is stored in inputs:
```
      "sshAuth": {
        "password": "AAABAD6/2Nroj62qORoHOLofFOkRhdUNwCAYeC86nABU/G4AO5I7Fw==",
        "sshPrivateKey": "AAABABqRIQ1bZbvU/hrlpX1Rh9sj9OCyArjG0SUILPQmb0KSCFIrz6bK"
      }
```

Passwords and sshKey are forced into twin secrets, Environment Variables are optionally twin secrets, everything else uses normal Pulumi workflows, because they are not secret in Pulumi Service.

Import of the above code generates successfully with dummy values for secrets:
```
const ds1 = new pulumiservice.DeploymentSettings("ds1", {
    operationContext: {
        environmentVariables: {
            SECRET_VAR: pulumi.secret("<REPLACE WITH ACTUAL SECRET VALUE>"),
            TEST_VAR: "fooa",
        },
    },
    organization: "IaroslavTitov",
    project: "PulumiDotnet",
    sourceContext: {
        git: {
            branch: "refs/heads/main",
            gitAuth: {
                sshAuth: {
                    password: pulumi.secret("<REPLACE WITH ACTUAL SECRET VALUE>"),
                    sshPrivateKey: pulumi.secret("<REPLACE WITH ACTUAL SECRET VALUE>"),
                },
            },
            repoDir: "pulumi-programs/simple-resource",
            repoUrl: "https://github.com/pulumi/deploy-demos.git",
        },
    },
    stack: "SdkTest5",
}, {
    protect: true,
});
```


provider/pkg/provider/deployment_settings.go
```diff
@@ -14,14 +14,16 @@ import (
...
+
 type PulumiServiceDeploymentSettingsInput struct {
 	pulumiapi.DeploymentSettings
 	Stack pulumiapi.StackName
 }
 
-const FixMe = "<value is secret and must be replaced>"
-
-func (ds *PulumiServiceDeploymentSettingsInput) ToPropertyMap() resource.PropertyMap {
+func (ds *PulumiServiceDeploymentSettingsInput) ToPropertyMap(plaintextSettings *pulumiapi.DeploymentSettings, oldCipherSettings *pulumiapi.DeploymentSettings, isInput bool) resource.PropertyMap {
```
komalali, line 26: Can we rename these variables just so the logical choices we're making later on are clearer? From what I understand, `plaintextSettings` is actually `oldInputs`? And `oldCipherSettings` is `oldOuputs`?

```suggestion
func (ds *PulumiServiceDeploymentSettingsInput) ToPropertyMap(plaintextSettings *pulumiapi.DeploymentSettings, oldCipherSettings *pulumiapi.DeploymentSettings, isInput bool) resource.PropertyMap {
```

provider/pkg/provider/deployment_settings.go
```diff
@@ -52,10 +54,47 @@ func (ds *PulumiServiceDeploymentSettingsInput) ToPropertyMap() resource.Propert
...
+							importSecretValue(sshAuthPropertyMap, "sshPrivateKey", ds.SourceContext.Git.GitAuth.SSHAuth.SSHPrivateKey, isInput)
+						} else {
+							if oldCipherSettings == nil {
+								createSecretValue(sshAuthPropertyMap, "sshPrivateKey", ds.SourceContext.Git.GitAuth.SSHAuth.SSHPrivateKey,
+									plaintextSettings.SourceContext.Git.GitAuth.SSHAuth.SSHPrivateKey, isInput)
+							} else {
+								if oldCipherSettings.SourceContext != nil &&
+									oldCipherSettings.SourceContext.Git != nil &&
+									oldCipherSettings.SourceContext.Git.GitAuth != nil &&
+									oldCipherSettings.SourceContext.Git.GitAuth.SSHAuth != nil {
```
komalali, line 67: Really wish we didn't have to do this :(

provider/pkg/provider/deployment_settings.go
```diff
@@ -482,6 +579,8 @@ func getSecretOrStringValue(prop resource.PropertyValue) string {
 	switch prop.V.(type) {
 	case *resource.Secret:
 		return prop.SecretValue().Element.StringValue()
+	case nil:
```
komalali, line 582: I don't think the type would ever be `nil`?

provider/pkg/provider/deployment_settings.go
```diff
@@ -577,33 +676,62 @@ func (ds *PulumiServiceDeploymentSettingsResource) Read(req *pulumirpc.ReadReque
 		DeploymentSettings: *settings,
 	}
 
+	var plaintextSettings *pulumiapi.DeploymentSettings = nil
+	var ciphertextSettings *pulumiapi.DeploymentSettings = nil
```
komalali, line 680: convention is to just leave these uninitiated

```suggestion
	var plaintextSettings *pulumiapi.DeploymentSettings
	var ciphertextSettings *pulumiapi.DeploymentSettings
```

provider/pkg/provider/deployment_settings.go
IaroslavTitov: Hmm, I think we should call objects of type DeploymentSettings `somethingSettings`. I don't think the method needs to know where these come from, just that one has plaintext in them and other has previous ciphertext. Maybe I can rename `old` in `oldCipherSettings` into `previous` or `last`, would that help? Or even `currentCiphertext`, to align closer with the [design](https://docs.google.com/document/d/1msoP-xNPGKOxGzTfvtYSANY_6T0NtiJHNpj6l5A2Y9g/edit#heading=h.qocf1eqtb2j9)

provider/pkg/provider/deployment_settings.go
IaroslavTitov: Fully agree, I hated every second of typing this out, but no way around in go that I know of...
I really miss `?.`

provider/pkg/provider/deployment_settings.go
IaroslavTitov: So this was actually a weird bug that I still don't quite understand, but sometimes it happens to be nil, it took me a couple hours to track down a panic to this, because I couldn't believe it. I added this dummy return hoping to figure out how it happens but nothing broke with the "" return at all, so I figured I'll leave it be. I can file an issue to track it down if you think it's worth it

provider/pkg/provider/deployment_settings.go
IaroslavTitov: Ack, will update, I just personally like express defaults

provider/pkg/provider/deployment_settings.go
IaroslavTitov: I can also add booleans up top to avoid repeating check of `plaintextSettings == nil`, I think that would make it more clear

komalali: Yeah this function was already pretty unwieldy to start with and I find it impossible to reason through some of the decisions we're making now. To me, it doesn't matter what has plaintext or what has ciphertext. The decisions are based on whether or not previous inputs or outputs are present or not. 

Honestly not sure what would make it better, I can understand it's subjective what makes more sense to me vs what makes more sense to you. At the very least can you add comments to explain what the code is doing?

provider/pkg/provider/deployment_settings.go
```diff
@@ -52,10 +57,47 @@ func (ds *PulumiServiceDeploymentSettingsInput) ToPropertyMap() resource.Propert
...
+									currentStateCipherSettings.SourceContext.Git.GitAuth.SSHAuth != nil &&
+									currentStateCipherSettings.SourceContext.Git.GitAuth.SSHAuth.Password != nil {
+									mergeSecretValue(sshAuthPropertyMap, "password", *ds.SourceContext.Git.GitAuth.SSHAuth.Password,
+										*plaintextInputSettings.SourceContext.Git.GitAuth.SSHAuth.Password,
+										*currentStateCipherSettings.SourceContext.Git.GitAuth.SSHAuth.Password, isInput)
+								} else {
+									importSecretValue(sshAuthPropertyMap, "password", *ds.SourceContext.Git.GitAuth.SSHAuth.Password, isInput)
+								}
+							}
+						}
```
komalali, line 100: Can you explain why multiple paths result in `importSecretValue`? Why are we importing if we're not in import mode?

provider/pkg/provider/deployment_settings.go
IaroslavTitov: Yeah, agree this function is a mess, I can maybe refactor it later and split it up.
`To me, it doesn't matter what has plaintext or what has ciphertext.` heh, well you kinda need to know that to extract the right data

Sounds good, I will add comments on how the parameters are used and what they mean

provider/pkg/provider/deployment_settings.go
IaroslavTitov: For sure, and I will also add it as a comment
So if both current state and newly read state have some value, we normally are in merge mode (checking ciphertexts and deciding whether we need to clean plaintext value)
BUT
If new state retrieved has a new value, but old ciphertext does NOT have that value, that means we need to just import it

I was trying to rephrase that logic eloquently to avoid two branches importing, but it was even more ugly than now... I'll take one more stab at it

komalali: > If new state retrieved has a new value, but old ciphertext does NOT have that value, that means we need to just import it

When does this happen?

provider/pkg/provider/deployment_settings.go
```diff
@@ -52,10 +66,43 @@ func (ds *PulumiServiceDeploymentSettingsInput) ToPropertyMap() resource.Propert
...
 					if ds.SourceContext.Git.GitAuth.SSHAuth.SSHPrivateKey.Value != "" {
-						sshAuthPropertyMap["sshPrivateKey"] = resource.NewPropertyValue(FixMe)
+						if mergeMode {
+							var plaintextValue *pulumiapi.SecretValue
+							var currentCipherValue *pulumiapi.SecretValue
+							if currentStateCipherSettings.SourceContext != nil &&
+								currentStateCipherSettings.SourceContext.Git != nil &&
+								currentStateCipherSettings.SourceContext.Git.GitAuth != nil &&
+								currentStateCipherSettings.SourceContext.Git.GitAuth.SSHAuth != nil {
+								plaintextValue = &plaintextInputSettings.SourceContext.Git.GitAuth.SSHAuth.SSHPrivateKey
```
komalali, line 76: Is this guaranteed to exist? What happens if `plaintextInputSettings.SourceContext.Git.GitAuth.SSHAuth` is `nil`?

IaroslavTitov: Yes, if a value in plaintextInputSettings exists, it also exists in plaintextInputSettings, because they are inputs and outputs of the current state

komalali: I'm guessing you meant "if a value in currentStateCipherSettings exists, it also exists in plaintextInputSettings"?

komalali (approved):

provider/pkg/provider/deployment_settings.go
IaroslavTitov: Ah, yes, sorry!

Issue #319, opened 2024-06-12
Cannot get an existing Environment


desteves:
### What happened?

I'm attempting to get an existing Environment using the `pulumiservice.Environment` resource but I get `error: Preview failed: failed to unmarshal organization value from properties: map[]`

### Example

`index.ts`

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as pulumiservice from "@pulumi/pulumiservice";


// pulumi config set envName aws-oidc-env

const config = new pulumi.Config();
const org = pulumi.getOrganization();
const envName = config.require("envName");
const envId =  `${org}/${envName}`;

const envRef =  pulumiservice.Environment.get(envId, envId);
```

### Output of `pulumi about`

CLI          
Version      3.119.0
Go Version   go1.22.3
Go Compiler  gc

Plugins
KIND      NAME           VERSION
language  nodejs         unknown
resource  pulumiservice  0.21.2

Host     
OS       darwin
Version  14.4.1
Arch     arm64

This project is written in nodejs: executable='/opt/homebrew/bin/node' version='v22.2.0'

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

IaroslavTitov:
I have reproduced the issue, but it so happened to be solved in https://github.com/pulumi/pulumi-pulumiservice/pull/326, which I was working on today. I will keep the issue open until I cut a new release, after that updating to new version will fix the issue.

IaroslavTitov:
The new release containing a fix is out - 0.21.4, resolving

PR #331, opened 2024-06-24
Codefreeze for 0.21.4

IaroslavTitov:


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+## 0.21.4
+
+### Miscellaneous
+
+- Fixed integ tests [#328](https://github.com/pulumi/pulumi-pulumiservice/issues/328)
```
komalali, line 8: I think the result was that this was not just fixing the tests? There were meaningful bug fixes included in this change. Can we please capture what we fixed for end users?

IaroslavTitov: Updated to include bug https://github.com/pulumi/pulumi-pulumiservice/issues/319, which was also fixed during the PR that fixed integ tests.

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
+- Fixed Environment Get function by fixing the resource's Read method [#319](https://github.com/pulumi/pulumi-pulumiservice/issues/319)
```
komalali, line 9: Should this be under `Bug Fixes`?

komalali (approved):

CHANGELOG.md
IaroslavTitov: Ah, you're right, I just added below without much thought

PR #334, opened 2024-06-25
Fix our deprecated actions usage

komalali:


IaroslavTitov (approved):
Thank you for bumping these up!

PR #336, opened 2024-06-25
Update another deprecated github action

komalali:


IaroslavTitov (approved):
Is this just a rename or does this newer action do anything else? `actions` is not super descriptive

komalali:
@IaroslavTitov it's this: https://github.com/pulumi/actions#installation-only

Which is the new version of this: https://github.com/pulumi/setup-pulumi

PR #337, opened 2024-06-25
Pin Twine to a good version

komalali:
Python SDK publishing broke with the latest version of importlib. See https://github.com/pulumi/pulumi-pulumiservice/actions/runs/9670567679/job/26679789077#step:21:216

See https://github.com/pulumi/ci-mgmt/pull/999/files


blampe (approved):
You might want to consider putting this provider under ci-mgmt so it can maintain these workflows for you.

komalali:
Opened https://github.com/pulumi/pulumi-pulumiservice/issues/338 to add this repo to ci-mgmt so we don't have to maintain these workflows ourselves

PR #341, opened 2024-06-28
Refactoring AccessToken Read method

IaroslavTitov:
### Summary
- `tokenId` property is not used anywhere at all, lol. The output tokenId is used as the resourceId, making this one useless
- Eliminated use of req.GetProperties in AccessToken, which fixed import

### Testing 
- pulumi import pulumiservice:index:AccessToken hui a7a15253-0108-4e39-89ff-3cd8859adba5

### DIFF
--- a/provider/cmd/pulumi-resource-pulumiservice/schema.json
+++ b/provider/cmd/pulumi-resource-pulumiservice/schema.json
@@ -564,10 +564,6 @@
     "pulumiservice:index:AccessToken": {
       "description": "Access tokens allow a user to authenticate against the Pulumi Cloud",
       "properties": {
-        "tokenId": {
-          "description": "The token identifier.",
-          "type": "string"
-        },
         "description": {
           "description": "Description of the access token.",
           "type": "string"
@@ -579,7 +575,6 @@
         }
       },
       "required": [
-        "tokenId",
         "description",
         "value"
       ],
--- a/provider/pkg/provider/access_token.go
+++ b/provider/pkg/provider/access_token.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	pbempty "google.golang.org/protobuf/types/known/emptypb"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/pulumi/pulumi-pulumiservice/provider/pkg/internal/pulumiapi"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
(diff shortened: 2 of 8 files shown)


provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -564,10 +564,6 @@
     "pulumiservice:index:AccessToken": {
       "description": "Access tokens allow a user to authenticate against the Pulumi Cloud",
       "properties": {
-        "tokenId": {
-          "description": "The token identifier.",
-          "type": "string"
-        },
```
komalali, line 570: Can you explain why we're removing this?

komalali: > The output tokenId is used as the resourceId, making this one useless

I see this in the description, I'm not sure I understand the implication.

IaroslavTitov: I mentioned in description:
- tokenId property is not used anywhere at all, lol. The output tokenId is used as the resourceId, making this one useless

komalali: Yeah, as I mentioned in the thread, I don't really understand what that's supposed to mean. Why is it useless? What does it mean that it's not really used anywhere?

provider/pkg/provider/access_token.go
```diff
@@ -20,10 +21,31 @@ type PulumiServiceAccessTokenInput struct {
...
-func (i *PulumiServiceAccessTokenInput) ToPropertyMap() resource.PropertyMap {
-	pm := resource.PropertyMap{}
-	pm["description"] = resource.NewPropertyValue(i.Description)
-	return pm
+// AccessToken uses outdated way of storing input in internal __inputs property
+func GenerateProperties(input PulumiServiceAccessTokenInput, accessToken pulumiapi.AccessToken) (outputs *structpb.Struct, inputs *structpb.Struct, err error) {
+	inputMap := resource.PropertyMap{}
+	inputMap["description"] = resource.NewPropertyValue(input.Description)
+	outputStore := resource.PropertyMap{}
+	outputStore["__inputs"] = resource.NewObjectProperty(inputMap)
```
komalali, line 29: Should we stop doing this `__inputs` thing like we did for webhooks: https://github.com/pulumi/pulumi-pulumiservice/pull/87

As described in the PR, [we don't do this __inputs thing in ay other providers](https://github.com/pulumi/pulumi-pulumiservice/pull/87#:~:text=We%20don%27t%20do%20this%20__inputs%20thing%20in%20any%20other%20providers) and have also been refactoring our code to move away from it in this repo.

provider/cmd/pulumi-resource-pulumiservice/schema.json
komalali: I'll ask the question a different way - what do we gain from making this breaking change?

IaroslavTitov: Oh, sorry, didn't update the page and never saw second comment ':D

So the property is literally not used. It's never filled out on create or update, never read from properties or anything. I think it was added originally, then author of AccessToken pivoted to storing tokenId in the resourceId and never cleaned up. That's just a guess as to why though.

We can leave it in to avoid breaking change, no problem with that really, except if a customer goes and tried to access the property and it's empty. Alternatively I can actually fill it out, but that would be storing the same data in 2 places.

So I think options are:
- remove (my preference, breaking change, but not going to impact customers)
- fill (redundant data storing)
- ignore (do nothing, leave it as empty property)

provider/pkg/provider/access_token.go
IaroslavTitov: I can do that, was trying to keep refactor just to the Read method fix

provider/cmd/pulumi-resource-pulumiservice/schema.json
komalali: Thanks for explaining! Agree with you, makes sense to remove it.

provider/pkg/provider/access_token.go
komalali: Totally cool to leave as-is for now, but let's open an issue to follow-up?

komalali (approved):

provider/pkg/provider/access_token.go
IaroslavTitov: Created https://github.com/pulumi/pulumi-pulumiservice/issues/342

PR #348, opened 2024-07-03
Fixing noisy diff in DS OIDC object

IaroslavTitov:
### Summary
- Fixed diff in azure object by actually adding it to the call - it was missed accidentally
- Fixed diff in aws object's duration field by normalizing the string in Check method, the same way Pulumi Service does (conversion into time.Duration and back into string)
- Removing SkipRefresh flag, now that DeploymentSettings are fully without diff

### Testing
- Manual test using TS program
- Integ tests

### DIFF
--- a/provider/pkg/provider/deployment_settings.go
+++ b/provider/pkg/provider/deployment_settings.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"time"
 
 	pbempty "google.golang.org/protobuf/types/known/emptypb"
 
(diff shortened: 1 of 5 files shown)


komalali (approved):

provider/pkg/provider/deployment_settings.go
```diff
@@ -236,6 +237,7 @@ func (ds *PulumiServiceDeploymentSettingsInput) ToPropertyMap(plaintextInputSett
 					if ds.OperationContext.OIDC.Azure.SubscriptionID != "" {
 						azureMap["subscriptionId"] = resource.NewPropertyValue(ds.OperationContext.OIDC.Azure.SubscriptionID)
 					}
+					oidcMap["azure"] = resource.PropertyValue{V: azureMap}
```
komalali, line 240: hehe whoops

PR #353, opened 2024-07-17
Fix Read method of EnvVersionTag

IaroslavTitov:
### Summary
- Stopped using req.GetProperties()

### Testing
- Verified import now works


provider/pkg/provider/environment_version_tags.go
```diff
@@ -189,3 +192,12 @@ func (evt *PulumiServiceEnvironmentVersionTagResource) Read(req *pulumirpc.ReadR
...
 }
+
+func splitEnvironmentTagId(id string) (string, string, string, error) {
+	// format: organization/environment/tag
+	s := strings.Split(id, "/")
+	if len(s) != 3 {
+		return "", "", "", fmt.Errorf("%q is invalid, must be in the format: organization/environment/tag", id)
+	}
+	return s[0], s[1], s[2], nil
+}
```
komalali, line 203: Could be nice to have this return a struct with Organization, Environment, Tag properties so it's not as dependent on knowing the order when you're consuming the output of the function.

komalali (approved):

PR #357, opened 2024-07-21
Stack Resource

### LINKED ISSUE #171: Create a new resource `stack`
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

When running `pulumi new ...` or `pulumi stack init`, it would be helpful if this provider could create new empty stacks and manage the stack through this provider.

komalali:
Create a Stack resource, so users can manage stacks within their pulumi program. This can enable, for example, the ability to ascribe naming conventions to stacks, or to create a stack with deployment settings and a schedule to regularly run updates, etc.

Tested this by updating one of the existing deployment settings tests to use this resource as the underlying stack. Also tested `import` and `refresh` manually.

Fixes #171


provider/pkg/internal/pulumiapi/stack_test.go
```diff
@@ -0,0 +1,74 @@
...
+	})
+}
+
+func TestDeleteStack(t *testing.T) {
+	s := StackIdentifier{
+		OrgName:     "organization",
+		ProjectName: "project",
+		StackName:   "stack",
+	}
+	t.Run("Happy Path", func(t *testing.T) {
```
seanyeh, line 52: may be good to add a test with forceDestroy true just to be complete

seanyeh (approved):

PR #370, opened 2024-08-06
Fixing exposed secret in webhook preview

IaroslavTitov:
### Summary
Added `KeepSecrets` where it was missing in marshaling logic, to make sure Webhook preview doesn't expose secrets.

### Testing
- Manually ran `pulumi up` -> details and verified it no longer shows up

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,6 @@
 ### Improvements
 
 ### Bug Fixes
+- Fixing webhook exposing secret values bug [#371](https://github.com/pulumi/pulumi-pulumiservice/issues/371)
 
 ### Miscellaneous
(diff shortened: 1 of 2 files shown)


komalali (commented):
Thanks for tackling this! Can we please check that:

- Everything deserializes correctly? For deployment settings I had to add this function to handle secrets and strings: https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/deployment_settings.go#L357
- We audit and open issues for other resources?

IaroslavTitov:
> * Everything deserializes correctly? For deployment settings I had to add this function to handle secrets and strings: https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/deployment_settings.go#L357

Verified that refresh works fine (integ tests double checked me). The secret is not returned, so no issues with deserializing that.

> * We audit and open issues for other resources?
I tested other resources by taking all of out yaml examples, gluing together into one program, then replacing all strings with ${mySecret}. Ran pulumi preview -> details and the secret did not show up as plaintext, so this is not an issue for other resources.

PR #375, opened 2024-08-13
Improving UX in Environment resource

IaroslavTitov:
### Summary:
- Add validation on Environment resource's values, blocking values using /
- Improve error messages of Environment resource
- Moved diags check above error, as it should precede (even though diags don't work currently, see [bug](https://github.com/pulumi/esc/issues/370))

### Testing
- Manually
- Integ tests


provider/pkg/provider/environment.go
```diff
@@ -211,6 +211,13 @@ func (st *PulumiServiceEnvironmentResource) Check(req *pulumirpc.CheckRequest) (
 				Reason:   fmt.Sprintf("missing required property '%s'", p),
 				Property: string(p),
 			})
+		} else {
+			if p != "yaml" && strings.Contains(inputMap[(p)].StringValue(), "/") {
```
seanyeh, line 215: NP `else if`?

seanyeh (approved):

PR #386, opened 2024-08-15
Adding TemplateSource resource

IaroslavTitov:
### Summary
- Adding TemplateSource resource, to have the ability to add templates to orgs via PSP

### Testing
- Manual test
- Added new integ tests


provider/pkg/provider/template_source.go
```diff
@@ -0,0 +1,280 @@
...
+		}, nil
+	}
+
+	dd := plugin.NewDetailedDiffFromObjectDiff(diffs, false)
+
+	detailedDiffs := map[string]*pulumirpc.PropertyDiff{}
+	replaces := []string(nil)
+	replaceProperties := map[string]bool{
+		"organizationName": true,
+	}
```
dschaller, line 94: just for my own knowledge why do we always replace `organizationName`

provider/pkg/internal/pulumiapi/template_source.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, fmt.Errorf("failed to get template source in org %s with id %s due to error: %w", organizationName, templateID, err)
+	}
+
+	for _, source := range templateSources.Sources {
+		if source.Id == templateID {
+			return &source, nil
+		}
+	}
+
+	return nil, nil
```
dschaller, line 69: not entirely sure of the pattern here but if we can't find a template source do we want to raise an error?

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1435,6 +1444,55 @@
...
+        "sourceName": {
+          "description": "Source name.",
+          "type": "string"
+        },
+        "sourceURL": {
+          "description": "Github URL of the repository from which to grab templates.",
+          "type": "string"
+        },
+        "destination": {
+          "description": "Settings that get filled in on new project creation.",
```
dschaller, line 1487:
```suggestion
          "description": "The default destination for projects using templates from this source.",
```

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1435,6 +1444,55 @@
...
```
dschaller, line 1464:
```suggestion
          "description": "The default destination for projects using templates from this source.",
```

provider/pkg/provider/template_source.go
IaroslavTitov: So if `organizationName` changes, we *can't* update the resource, like, a TemplateSource created in one org, cannot be moved into another org.
So we have to delete old one and make new one in new org

provider/pkg/internal/pulumiapi/template_source.go
IaroslavTitov: We don't want to throw an error here, for example during refresh we expect it to be not found (if user deleted the template source, causing a drift)

I mean, I could refactor this to return NotFound error here instead and have caller handle that, but imho this way is better, since there was no error calling API, just nothing (nil) found

provider/cmd/pulumi-resource-pulumiservice/schema.json
IaroslavTitov: Thank you, I was struggling to come up with a good description here!

provider/pkg/internal/pulumiapi/template_source.go
```diff
@@ -0,0 +1,82 @@
...
+	_, err := c.do(ctx, http.MethodPatch, apiPath, request, &response)
+	if err != nil {
+		return nil, fmt.Errorf("failed to update template source in org %s with id %s: %+v due to error: %w", organizationName, templateId, request, err)
+	}
+	return &response, nil
+}
+
+func (c *Client) GetTemplateSource(ctx context.Context, organizationName string, templateID string) (*TemplateSourceResponse, error) {
+	// This sucks, but there's not Get API for Template Sources
+	// Thus, using a List and then finding by ID
```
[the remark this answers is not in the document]
IaroslavTitov: Added https://github.com/pulumi/pulumi-service/issues/21637, good point!

provider/pkg/provider/template_source.go
dschaller: ah ok - sorry the naming here threw me. These are properties that will trigger a replacement requirement no properties to replace.

PR #396, opened 2024-08-26
Add a python test for agent pools

komalali:


seanyeh (approved):
Looks fine to me. Do we want to add agent pool examples for other languages?

komalali:
@seanyeh I added one for YAML earlier and added this one because the reported case was specifically a Python example. I think just these are good enough for now.

PR #398, opened 2024-08-27
(WIP) Add ESC Projects

seanyeh:

PR #406, opened 2024-09-10
Adding environment webhook support, webhook groups

IaroslavTitov:
### Summary 
- Adding webhook group support
- Adding environment webhook support
- I have to build this on top on environment project support PR, so this will need to go in only after [this](https://github.com/pulumi/pulumi-pulumiservice/pull/398)

### Testing
- Manual test
- Updating example/integ test


provider/pkg/provider/webhook.go
```diff
@@ -424,15 +496,18 @@ func (wh *PulumiServiceWebhookResource) getWebhook(id string) (*pulumiapi.Webhoo
...
 	}
 	return webhook, nil
 }
 
 func splitWebhookID(id string) (*webhookID, error) {
-	// format: organization/project/stack/webhookName (stack webhook) or organization/webhookName (org webhook)
+	// format:
+	// organization/project/stack/webhookName (stack webhook)
+	// organization/webhookName (org webhook)
+	// organization/environment/project/stack/webhookName (stack webhook)
```
seanyeh, line 510: this should be env

seanyeh (approved):
looks good besides comment

provider/pkg/provider/webhook.go
IaroslavTitov, line 510:
```suggestion
	// organization/environment/projectName/environmentName/webhookName (environment webhook)
```

IaroslavTitov: Good catch, ty!

provider/pkg/provider/webhook.go
```diff
@@ -238,6 +301,10 @@ func (wh *PulumiServiceWebhookResource) createWebhook(input PulumiServiceWebhook
 	if input.ProjectName != nil && input.StackName != nil {
 		hookID = fmt.Sprintf("%s/%s/%s/%s", input.OrganizationName, *input.ProjectName, *input.StackName,
 			webhook.Name)
+	} else if input.ProjectName != nil && input.EnvironmentName != nil {
+		// This is not ideal, but inserting "environment" string to distinguish from stack webhooks
+		hookID = fmt.Sprintf("%s/%s/%s/%s/%s", input.OrganizationName, "environment", *input.ProjectName, *input.EnvironmentName,
+			webhook.Name)
```
dschaller, line 307:
```suggestion
		hookID = fmt.Sprintf("%s/environment/%s/%s/%s", input.OrganizationName, *input.ProjectName, *input.EnvironmentName,
			webhook.Name)
```

provider/pkg/provider/webhook.go
dschaller:
```suggestion
	// organization/environment/project/env/webhookName (stack webhook)
```

provider/pkg/provider/webhook.go
```diff
@@ -424,15 +496,18 @@ func (wh *PulumiServiceWebhookResource) getWebhook(id string) (*pulumiapi.Webhoo
...
 	if err != nil {
 		return nil, err
 	}
 	return webhook, nil
 }
 
 func splitWebhookID(id string) (*webhookID, error) {
-	// format: organization/project/stack/webhookName (stack webhook) or organization/webhookName (org webhook)
+	// format:
+	// organization/project/stack/webhookName (stack webhook)
```
dschaller, line 508: should we be consistent here and have the identifier always prefixed onto ID (e.g. stack vs env vs org) instead of inferring

provider/pkg/provider/webhook.go
IaroslavTitov: Sean spotted the same thing, fixed!

provider/pkg/provider/webhook.go
IaroslavTitov: Unfortunately we have resources already created with those id's, so any format change would be breaking. I wish it could be consistent as well :c

sdk/go/pulumiservice/pulumiEnums.go
```diff
@@ -816,6 +1064,9 @@ func init() {
 	pulumi.RegisterInputType(reflect.TypeOf((*WebhookFiltersArrayInput)(nil)).Elem(), WebhookFiltersArray{})
 	pulumi.RegisterInputType(reflect.TypeOf((*WebhookFormatInput)(nil)).Elem(), WebhookFormat("raw"))
 	pulumi.RegisterInputType(reflect.TypeOf((*WebhookFormatPtrInput)(nil)).Elem(), WebhookFormat("raw"))
+	pulumi.RegisterInputType(reflect.TypeOf((*WebhookGroupInput)(nil)).Elem(), WebhookGroup("stacks"))
+	pulumi.RegisterInputType(reflect.TypeOf((*WebhookGroupPtrInput)(nil)).Elem(), WebhookGroup("stacks"))
```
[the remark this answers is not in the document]
IaroslavTitov: SDKs are auto-generated. My guess here is that the first field is used as the default. Right above, "raw" is used, even though WebhookFormat also has many values

Issue #411, opened 2024-09-20
panic in `pulumiservice.Environment.yaml`  v0.26


desteves:
### What happened?

The latest version of the pulumi service seems to have a bug with the Environment yaml field.

Using `v0.23.2` does not have this potential bug as the resource is successfully created/previewed. See below for the error message encountered.

### Example

Example code:

- https://github.com/pulumi/examples/blob/master/aws-ts-oidc-provider-pulumi-cloud/index.ts#L86
- Another example: https://github.com/pulumi/workshops/blob/main/gitlab/pulumi-cloud-gitlab-integration/index.ts#L88

```bash

pulumi preview

## output

     └─ pulumiservice:index:Environment  aws-oidc-admin             1 error

Diagnostics:
  pulumi:pulumi:Stack (PROJECT-q):
    panic: interface conversion: interface {} is resource.Computed, not *asset.Asset
    goroutine 40 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/resource.PropertyValue.AssetValue(...)
      /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.131.0/go/common/resource/properties.go:456
    github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*PulumiServiceEnvironmentResource).Check(0x1289f40?, 0xc000450fc0?)
      /home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/environment.go:241 +0x778
    github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*pulumiserviceProvider).Check(0xc000450240, {0xc00041b720?, 0x4107e5?}, 0x40?)
      /home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/provider.go:217 +0xdb
    github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Check_Handler.func1({0x1298c18, 0xc0004c4bd0}, {0x100b2c0?, 0xc000450fc0})
      /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.131.0/proto/go/provider_grpc.pb.go:615 +0x72
[...]

PR #414, opened 2024-09-23
Fix panic when environment uses computed values

seanyeh:
Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/411

This is related to https://github.com/pulumi/pulumi-pulumiservice/pull/399

Example repro program:
```typescript
import * as pulumi from "@pulumi/pulumi";
import * as service from "@pulumi/pulumiservice";
import * as aws from "@pulumi/aws";

const org = "big-org";
const bucket = new aws.s3.Bucket("myBucket", {
    bucket: "syeh-panic-test",
    acl: "private",
});

const envDef = pulumi.interpolate`values:
  myKey1: "${bucket.arn}"`;
const envAsset = envDef.apply(x => new pulumi.asset.StringAsset(x));

const environment = new service.Environment("panic", {
  organization: org,
  project: "abc",
  name: "panic",
  yaml: envAsset,
})
```

### DIFF
--- a/provider/pkg/provider/environment.go
+++ b/provider/pkg/provider/environment.go
@@ -238,10 +238,14 @@ func (st *PulumiServiceEnvironmentResource) Check(req *pulumirpc.CheckRequest) (
 		}
 	}
 
-	yamlBytes, err := getBytesFromAsset(inputMap["yaml"].AssetValue())
-	if err != nil {
-		return nil, err
+	yamlBytes := []byte{}
+	if !inputMap["yaml"].IsComputed() {
+		yamlBytes, err = getBytesFromAsset(inputMap["yaml"].AssetValue())
+		if err != nil {
+			return nil, err
+		}
 	}
+
 	stringYaml := string(yamlBytes)
 	trimmedYaml := strings.TrimSpace(stringYaml)
 	inputMap["yaml"] = resource.MakeSecret(resource.NewStringProperty(trimmedYaml))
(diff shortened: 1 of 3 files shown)


provider/pkg/provider/environment.go
```diff
@@ -238,10 +238,14 @@ func (st *PulumiServiceEnvironmentResource) Check(req *pulumirpc.CheckRequest) (
 		}
 	}
 
-	yamlBytes, err := getBytesFromAsset(inputMap["yaml"].AssetValue())
-	if err != nil {
-		return nil, err
+	yamlBytes := []byte{}
```
komalali, line 241: Hmm... does this mean there will be no `yamlBytes` if it's computed? Or does this only affect `Check` and the resolved value of environment.yaml is still correct?

seanyeh: This only affects `Check`. I have confirmed the resolved value is still correct (I'm assuming that it's ok to not have computed values during Check?)

komalali: The purpose of this Check is to take the string value of the yaml and put that into state, so there's always a string value to compare against. In this case, when it's computed, it'll result in yaml being empty in state. So next time there's an update, there will be a diff because there is no value for `yaml`. You can test this by adding an example test similar to in https://github.com/pulumi/pulumi-pulumiservice/pull/391/files

We'll need a way to get the string value even out of the computed property I think.

pgavlin: Computed/unknown values will only appear during a preview, so I think that this will be okay. We'll re-run the `Check` during the update, at which point all values will be known.

pgavlin: Though I do think I'd do this slightly differently--instead of checking `IsAsset` I'd check for `!IsComputed`.

pgavlin (approved):
can we add a test for this (do we have tests for the provider in here in general)?

Issue #418, opened 2024-09-26
panic in the latest version of the pulumi service has a bug with the Environment yaml field.


firemanxbr:
### What happened?

The latest version of the pulumi service has a bug with the Environment yaml field.

```
Diagnostics:
  pulumiservice:index:Environment (gcp-non-prod):
    error: error reading from server: EOF

  pulumi:pulumi:Stack (iac-gcp-non-prod-green):
    panic: interface conversion: interface {} is string, not *asset.Asset
    goroutine 16 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/resource.PropertyValue.AssetValue(...)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.131.0/go/common/resource/properties.go:456
    github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*PulumiServiceEnvironmentResource).Check(0x10528efe8?, 0x14000156540?)
    	/home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/environment.go:243 +0x6f8
    github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider.(*pulumiserviceProvider).Check(0x1400004e600, {0x104b8d9c4?, 0x40?}, 0x1050a36e0?)
    	/home/runner/work/pulumi-pulumiservice/pulumi-pulumiservice/provider/pkg/provider/provider.go:217 +0xe4
    github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Check_Handler.func1({0x10529df98, 0x14000694360}, {0x1051ddc20?, 0x14000156540})
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.131.0/proto/go/provider_grpc.pb.go:615 +0x74
    github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc.OpenTracingServerInterceptor.func1({0x10529df98, 0x14000694180}, {0x1051ddc20, 0x14000156540}, 0x140003c8020, 0x14000494030)
    	/home/runner/go/pkg/mod/github.com/grpc-ecosystem/grpc-opentracing@v0.0.0-20180507213350-8e809c8a8645/go/otgrpc/server.go:57 +0x2e8
[...]

PR #421, opened 2024-10-02
Fix the yaml panic

komalali:
Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/418


seanyeh (approved):
just want to make sure we test this with both computed/non-computed values but otherwise 👍

seanyeh:
@komalali I tested this locally, confirmed fixes the panic and working with both computed/non-computed

PR #427, opened 2024-10-11
Make sourceContext optional

komalali:
In cases where a self-hosted runner is being used, the source code may be baked into the runner. In those cases, an additional source step is unnecessary.

### DIFF
--- a/examples/examples_yaml_test.go
+++ b/examples/examples_yaml_test.go
@@ -208,6 +208,43 @@ func TestYamlDeploymentSettingsExample(t *testing.T) {
 	})
 }
 
+func TestYamlDeploymentSettingsNoSourceExample(t *testing.T) {
+
+	// Set up tmpdir with a Pulumi.yml with no resources
+	// mimicking the deletion of resource
+	newProgram := YamlProgram{
+		Name:    "yaml-deployment-settings-example-no-source",
+		Runtime: "yaml",
+	}
+
+	tmpdir := writePulumiYaml(t, newProgram)
+
+	cwd, _ := os.Getwd()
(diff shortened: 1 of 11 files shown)


examples/yaml-deployment-settings/.pulumi
IaroslavTitov, line 1: What are the empty files for?

examples/examples_yaml_test.go
```diff
@@ -208,6 +208,43 @@ func TestYamlDeploymentSettingsExample(t *testing.T) {
 	})
 }
 
+func TestYamlDeploymentSettingsNoSourceExample(t *testing.T) {
+
+	// Set up tmpdir with a Pulumi.yml with no resources
```
IaroslavTitov, line 213: How does this work? Is the ProgramTest running `up` on the folder with Pulumi.yml and then on a non-existing folder? Why is this needed?

komalali: This is copy-paste from the other TestYamlDeploymentSettingsExample above.

examples/yaml-deployment-settings/.pulumi
komalali: They're just being deleted, there was an issue with the integration test framework earlier where it was forcing us to add these empty files but the bug was fixed.

examples/examples_yaml_test.go
IaroslavTitov: Duh, should've scrolled

Still curious as to why we need these temp dirs, but I can take a look later, unrelated to the change

IaroslavTitov (approved):

IaroslavTitov (approved):

PR #435, opened 2024-11-01
Allow force deleting agent pools

komalali:
Sometimes, a user may want to force delete an agent pool, even if it is associated with existing deployment settings. This is already supported in the API, bringing the functionality to PSP.

### DIFF
--- a/examples/ts-webhooks/index.ts
+++ b/examples/ts-webhooks/index.ts
@@ -39,7 +39,7 @@ const stackWebhook = new service.Webhook("stack-webhook", {
   organizationName: serviceOrg,
   projectName: pulumi.getProject(),
   stackName: pulumi.getStack(),
-  payloadUrl: "https://example.com",
+  payloadUrl: "https://hooks.slack.com/blahblah",
   format: WebhookFormat.Slack,
   groups: [ WebhookGroup.Stacks ],
   filters: [WebhookFilters.DeploymentStarted, WebhookFilters.DeploymentSucceeded],
--- a/provider/pkg/internal/pulumiapi/agent_pools.go
+++ b/provider/pkg/internal/pulumiapi/agent_pools.go
@@ -18,13 +18,14 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"path"
 )
 
 type AgentPoolClient interface {
 	CreateAgentPool(ctx context.Context, orgName, name, description string) (*AgentPool, error)
 	UpdateAgentPool(ctx context.Context, agentPoolId, orgName, name, description string) error
-	DeleteAgentPool(ctx context.Context, agentPoolId, orgName string) error
+	DeleteAgentPool(ctx context.Context, agentPoolId, orgName string, forceDestroy bool) error
 	GetAgentPool(ctx context.Context, agentPoolId, orgName string) (*AgentPool, error)
 }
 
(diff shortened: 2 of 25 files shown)


provider/pkg/provider/agent_pool_test.go
```diff
@@ -24,20 +30,75 @@ func (c *AgentPoolClientMock) CreateAgentPool(ctx context.Context, name, orgName
...
+		provider := PulumiServiceAgentPoolResource{
+			client: mockedClient,
+		}
+
+		req := pulumirpc.DeleteRequest{
+			Id:  "org/abc/beep-boop",
+			Urn: "urn:beep-boop",
+		}
+
+		resp, err := provider.Delete(&req)
```
nyobe, line 67: isn't this just testing that the mock returns what you told it to return? 😞
I don't think this is useful

examples/ts-webhooks/index.ts
```diff
@@ -39,7 +39,7 @@ const stackWebhook = new service.Webhook("stack-webhook", {
   organizationName: serviceOrg,
   projectName: pulumi.getProject(),
   stackName: pulumi.getStack(),
-  payloadUrl: "https://example.com",
+  payloadUrl: "https://hooks.slack.com/blahblah",
```
komalali, line 42: Needed an update due to new URL validation for webhooks

provider/pkg/provider/agent_pool_test.go
komalali: Yeah that's fair. Anything that's not a full integration test has been somewhat tricky to test in the provider. Let me spend some time figuring this out.

provider/pkg/internal/pulumiapi/agent_pools.go
```diff
@@ -106,18 +107,23 @@ func (c *Client) UpdateAgentPool(ctx context.Context, agentPoolId, orgName, name
 	return nil
 }
 
-func (c *Client) DeleteAgentPool(ctx context.Context, agentPoolId, orgName string) error {
+func (c *Client) DeleteAgentPool(ctx context.Context, agentPoolId, orgName string, forceDestroy bool) error {
```
dschaller, line 110: super nit: should we consider making options like this a struct in case we need to add more later?

provider/pkg/provider/agent_pool.go
```diff
@@ -31,6 +32,9 @@ func GenerateAgentPoolProperties(input PulumiServiceAgentPoolInput, agentPool pu
 	if input.Description != "" {
 		inputMap["description"] = resource.NewPropertyValue(input.Description)
 	}
+	if input.ForceDestroy {
+		inputMap["forceDestroy"] = resource.NewPropertyValue(input.ForceDestroy)
```
dschaller, line 36: Might want to make `forceDestroy` a constant

komalali: Hmm, not sure it makes sense to make one input a constant and not any of the others... so I'd leave that to a future optimization since this is not the pattern anywhere in the provider.

provider/pkg/provider/agent_pool_test.go
nyobe: it seems like most of the other tests in the package are the same way 🤷‍♀️ 
if you figure out something I'd love to know for #436 !

nyobe (approved):

provider/pkg/provider/agent_pool_test.go
komalali: Yeah I think I'm going to merge this just to unblock Scala but this is a good question that we do need an answer to beyond E2E tests

PR #436, opened 2024-11-06
Add support for DeploymentSettings dependency CacheOptions

nyobe:
Adds support for enabling dependency caching in deployment settings.


sdk/dotnet/DeploymentSettings.cs
IaroslavTitov, line 1: I think you missed a sdk-generated file, that's why wf is complaining. Should be as easy as `make build` `git add sdk` and re-push

IaroslavTitov (commented):
Overall looks good, would you please add CacheOptions into DeploymentSettings examples/integ tests?

sdk/dotnet/DeploymentSettings.cs
IaroslavTitov: Oh, also need an entry in CHANGELOG_PENDING

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -554,6 +554,17 @@
         }
       ]
     },
+    "pulumiservice:index:DeploymentSettingsCacheOptions": {
+      "description": "Dependency cache settings for the deployment",
+      "properties": {
+        "enabled": {
```
IaroslavTitov, line 560: Thanks for catching this in the docs PR!
Should be `enable`, without the d, to match the API

(I personally also prefer `enabled` but consistency is more important)

IaroslavTitov (approved):
Should change `enabled` -> `enable`, other than that LGTM

provider/cmd/pulumi-resource-pulumiservice/schema.json
nyobe: updated! https://github.com/pulumi/pulumi-pulumiservice/pull/436/commits/d019a21572b11d03d52c751545f386568f9bf07a

PR #455, opened 2024-12-06
enable passed one-time scheduled actions to be rescheduled

### LINKED ISSUE #449: pulumiservice.TtlSchedule can be created without pulumiservice.DeploymentSettings (and subsequent `up`s will error out)
### What happened?

It's possible to include a `pulumiservice.TtlSchedule` resource in a stack without a corresponding `pulumiservice.DeploymentSettings` resource. Running `pulumi up` with just the former will stand up the stack without any error. But then both 1) the TTL schedule will not _work_ and 2) subsequent `up`s will return an error like:

```
Diagnostics:
  pulumiservice:index:TtlSchedule (mystack-ttl-schedule):
    error: failed to update ttl schedule e29c5797-6c39-4477-9243-35f5eea2a90e (timestamp=2024-11-22 01:47:05 +0000 UTC, deleteAfterDestroy=true): 404 API error: Not Found: ScheduledAction 'e29c5797-6c39-4477-9243-35f5eea2a90e' not found

  pulumi:pulumi:Stack (mystack):
    error: update failed
```

There should be better error handling and messaging.

### Example

(see above)

### Output of `pulumi about`

```
CLI          
Version      3.140.0
Go Version   go1.23.3
Go Compiler  gc

Plugins
KIND      NAME             VERSION
resource  aws              6.55.0
resource  awsx             2.16.1
resource  datadog          4.34.0
resource  docker           4.5.6
resource  postgresql       3.13.0
resource  pulumi_aws_tags  1.2.1
resource  pulumiservice    0.26.2
language  python           unknown
resource  rabbitmq         3.3.7
resource  random           4.16.7
resource  synced-folder    0.11.1

Host     
OS       darwin
Version  14.5
Arch     arm64

This project is written in python: executable='/Users/me/Library/Caches/pypoetry/virtualenvs/infra-jqilnyNa-py3.12/bin/python3' version='3.12.7'

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/yyy
User           yyy
Organizations  yyy, xxx
Token type     personal

Dependencies:
NAME                  VERSION
black                 24.10.0
pip                   24.2
pre_commit            4.0.1
psycopg               3.2.3
pulumi_aws_tags       1.2.1
pulumi_awsx           2.16.1
pulumi_datadog        4.34.0
pulumi_postgresql     3.13.0
pulumi_pulumiservice  0.26.2
pulumi_rabbitmq       3.3.7
pulumi_random         4.16.7
pulumi_synced_folder  0.11.1
pyright               1.1.385
ruff                  0.6.9
types-PyYAML          6.0.12.20240917

Pulumi locates its logs in /var/folders/p1/cwqtcbkn4xvdt377y9jzrb180000gn/T/ by default
warning: Failed to get information about the current stack: No current stack
```

### Additional context

_No response_

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

nyobe:
One-time schedules are [soft-deleted after running](https://github.com/pulumi/pulumi-service/blob/186e4d4249d635f8f928f416c371c739ae0b52b5/cmd/service/model/schedules.go#L457), which means updating their schedule doesn't work because the action is never actually picked up again by the scheduler. (Prior to https://github.com/pulumi/pulumi-service/pull/24133 the update would appear to succeed, but the scheduled action would just silently never run. Now at least you get an error indicating the schedule was deleted.)

This PR enables rescheduling to work correctly by forcing the schedule to be replaced when its timestamp is updated.

Partially addresses #449

### DIFF
--- a/provider/cmd/pulumi-resource-pulumiservice/schema.json
+++ b/provider/cmd/pulumi-resource-pulumiservice/schema.json
@@ -1254,7 +1254,8 @@
         },
         "timestamp": {
           "description": "The time at which the schedule should run, in ISO 8601 format. Eg: 2020-01-01T00:00:00Z. If you are supplying this, do not supply scheduleCron.",
-          "type": "string"
+          "type": "string",
+          "replaceOnChanges": true
         },
         "pulumiOperation": {
           "description": "Which operation to run.",
@@ -1291,7 +1292,8 @@
         },
         "timestamp": {
           "description": "The time at which the schedule should run, in ISO 8601 format. Eg: 2020-01-01T00:00:00Z. If you are supplying this, do not supply scheduleCron.",
-          "type": "string"
+          "type": "string",
+          "replaceOnChanges": true
         },
         "pulumiOperation": {
           "description": "Which command to run.",
@@ -1388,7 +1390,8 @@
         },
         "timestamp": {
           "description": "The time at which the schedule should run, in ISO 8601 format. Eg: 2020-01-01T00:00:00Z.",
-          "type": "string"
+          "type": "string",
+          "replaceOnChanges": true
         },
         "deleteAfterDestroy": {
           "description": "True if the stack and all associated history and settings should be deleted.",
@@ -1422,7 +1425,8 @@
         },
         "timestamp": {
           "description": "The time at which the schedule should run, in ISO 8601 format. Eg: 2020-01-01T00:00:00Z.",
-          "type": "string"
+          "type": "string",
+          "replaceOnChanges": true
         },
         "deleteAfterDestroy": {
           "description": "True if the stack and all associated history and settings should be deleted.",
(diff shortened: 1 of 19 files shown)


komalali (approved):

IaroslavTitov (approved):

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1254,7 +1254,8 @@
         },
         "timestamp": {
           "description": "The time at which the schedule should run, in ISO 8601 format. Eg: 2020-01-01T00:00:00Z. If you are supplying this, do not supply scheduleCron.",
-          "type": "string"
+          "type": "string",
+          "willReplaceOnChanges": true
```
komalali, line 1258: I don't think this is a real property that has any effect on the generated SDK. `replaceOnChanges` does, but we don't actually need it anymore since we're handling it in the provider code.

komalali, line 1258:
```suggestion
          "type": "string"
```

komalali: Ignore me!

komalali: Turns out `willReplaceOnChanges` is a [real thing](https://www.pulumi.com/docs/iac/packages-and-automation/pulumi-packages/schema/#:~:text=a%20update%20operation.-,willReplaceOnChanges,-boolean) used as documentation to indicate that the provider handles replacement.

Issue #477, opened 2024-12-16
Resource plugin version mismatch despite no conflicting binaries


aureq:
### What happened?

When I run `p up` or `p pre` and my Pulumi app uses the Pulumi Service provider, as of version 0.27.3, it returns the following warning

```
warning: resource plugin pulumiservice is expected to have version >=0.27.3, but has 0.0.1; the wrong version may be on your path, or this may be a bug in the plugin
```

Not affected: `0.27.1`
Affected: `0.27.2` and `0.27.3`

While the warning is annoying this doesn't seem to affect the management of resources.

### Example

```ts
import * as pulumi from "@pulumi/pulumi";
import * as pulumiservice from "@pulumi/pulumiservice";

export = async() => {

    const serviceName = "platform-infra";

    const org = pulumi.getOrganization();

    const escAwsAccess = new pulumiservice.Environment(`${serviceName}-oidc-provider`, {
        name: "oidc-access",
        organization: org,
        project: "aws",
        yaml: "",
    });

    return {

    };
};
```

### Output of `pulumi about`

```
CLI          
Version      3.143.0
Go Version   go1.23.4
Go Compiler  gc

Plugins
KIND      NAME           VERSION
resource  aws            6.64.0
resource  aws-native     1.14.0
language  nodejs         3.143.0
resource  pulumiservice  0.27.3

Host     
OS       debian
Version  12.8
Arch     x86_64

This project is written in nodejs: executable='/usr/local/bin/node' version='v22.11.0'

Current Stack: menfin-demo/platform-infra/prod

TYPE                             URN
pulumi:pulumi:Stack              urn:pulumi:prod::platform-infra::pulumi:pulumi:Stack::platform-infra-prod
pulumi:providers:pulumiservice   urn:pulumi:prod::platform-infra::pulumi:providers:pulumiservice::default_0_27_3
pulumiservice:index:Environment  urn:pulumi:prod::platform-infra::pulumiservice:index:Environment::platform-infra-oidc-provider


Found no pending operations associated with menfin-demo/prod

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/aureq
User           aureq
Organizations  aureq, team-ce, menfin-demo, menfin, menfin-team, demo
Token type     personal

Dependencies:
NAME                   VERSION
@pulumi/aws-native     1.14.0
@pulumi/aws            6.64.0
@pulumi/pulumi         3.142.0
@pulumi/pulumiservice  0.27.3
[...]

Issue #465, opened 2024-12-09
[Housekeeping] Eliminate useless files


IaroslavTitov:
### Summary
- While looking around planning for PSPV1, noticed files called version.go and serve.go
  - Both are very short and don't seem to need to be there
  - investigate and eliminate extra files if possible
- There is also a unkown.go file representing an unknown resource
  - Remove that file - we can throw errors directly

IaroslavTitov:
The first bullet point is done

PR #486, opened 2025-01-10
Fix passing in version on provider build

IaroslavTitov:
### Summary
- Fixes: https://github.com/pulumi/pulumi-pulumiservice/issues/477
- Fixes the first bullet point of https://github.com/pulumi/pulumi-pulumiservice/issues/465
- Fixing missed Version insertion using -ldflags
- Moving location of Version to main.go to eliminate empty file
- Moving serve.go logic to main.go to pull out serving logic out of provider package
- Minor cleanup of Makefile - creating one BUILD_PATH variable to avoid repeating path blocks
- Adding missed version specification for java sdk generation

### Summary
- Tested build works locally and provider works as expected
- Integ tests


t0yv0 (approved):
Not spotting any problems per se, but this will be confusing: when a repository is onboarded to ci-mgmt usually these files are auto-generated:

```
.github/workflows/build_sdk.yml
```

Makefile may or may not be generated - it depends but we need to be clear which one it is. Assuming this repo has a custom Makefile.

sdk/java/build.gradle is usually automatically generated as well by the make java_sdk command. 

The danger here is that if these are actually automatically generated, the manual changes might be lost next time this is done.

IaroslavTitov:
> build_sdk.yml

Is it supposed to be auto-generated every run or once upon onboarding? It's not regenerated every time today afaik
I had to edit that file because I added the missed `--version` into java sdk generation, so `sdk/java/build.gradle` is now generated with a different version string every time, and thus needed changes to be ignored

> we need to be clear which one it is. Assuming this repo has a custom Makefile.

This is a custom Makefile, never generated afaik. Should I add a comment or something like that to make it explicit that it is?

t0yv0:
Perfect thanks for clarifying. So Makefile is custom. sdk/java/build.gradle is generated - all good apparently.

This leaves .github/workflows/build_sdk.yml

This file should be coming from ci-mgmt and as it evolves in ci-mgmt this repo should be getting an updated copy. Usually this happens in automated PRs like https://github.com/pulumi/pulumi-aws/pull/5092 - this helps us avoid drift of these workflows across all providers.

IaroslavTitov:
> This file should be coming from ci-mgmt and as it evolves in ci-mgmt this repo should be getting an updated copy.

Ah, I see, good to know!

Issue #481, opened 2024-12-19
Standartize Diff methods


IaroslavTitov:
### Summary
- I noticed that Diff methods are very similar, the only difference being the replace properties pretty much
- Create a shared method for all resources to use
- Some of the replace logic can be handled by adding "replaceOnChange" in the schema? Test this and remove explicit logic if that's the case!

Issue #503, opened 2025-03-06
Replace hardcoded bool defaults in code with schema


IaroslavTitov:
### Ask
- Schedules hardcode default values in Diff method
- See if that can be replaced by using `default` in schema instead

Issue #342, opened 2024-06-28
Migrate all resources away from __input property


IaroslavTitov:
See https://github.com/pulumi/pulumi-pulumiservice/pull/87/files for reference. Find all resources that still use legacy __input and migrate them similar to the PR mentioned.

PR #504, opened 2025-03-06
Standartizing PSP diffs

IaroslavTitov:
### Summary
- Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/481
- Sadly, confirmed that schema param `replaceOnChanges` does NOT work as expected, so using hardcoded properties in the diff methods instead. (I suspect it will work once we move to new `go-provider`, can then get rid of these)
- Standartizing all the Diff methods to use the same logic, with a couple exceptions:
  - schedules need to set default bools (cut https://github.com/pulumi/pulumi-pulumiservice/issues/503 to followup)
  - webhooks and access tokens use deprecated `__input` way of storing inputs (we have https://github.com/pulumi/pulumi-pulumiservice/issues/342 to take care of that)
- Removes 100s of lines of redundant code

### DIFF
--- a/provider/cmd/pulumi-resource-pulumiservice/schema.json
+++ b/provider/cmd/pulumi-resource-pulumiservice/schema.json
@@ -1322,36 +1322,6 @@
         "project",
         "stack",
         "permission"
-      ],
-      "properties": {
-        "team": {
-          "description": "The name of the team to grant this stack permissions to. This is not the display name.",
-          "type": "string"
-        },
-        "organization": {
-          "description": "The organization or the personal account name of the stack.",
-          "type": "string"
-        },
-        "project": {
-          "description": "The project name for this stack.",
-          "type": "string"
-        },
-        "stack": {
-          "description": "The name of the stack that the team will be granted permissions to.",
-          "type": "string"
-        },
-        "permission": {
-          "$ref": "#/types/pulumiservice:index:TeamStackPermissionScope",
-          "plain": true,
-          "description": "Sets the permission level that this team will be granted to the stack."
-        }
-      },
-      "required": [
-        "team",
-        "organization",
-        "project",
-        "stack",
-        "permission"
       ]
     },
     "pulumiservice:index:DeploymentSchedule": {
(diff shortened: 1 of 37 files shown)


provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1345,8 +1375,7 @@
         },
         "timestamp": {
           "description": "The time at which the schedule should run, in ISO 8601 format. Eg: 2020-01-01T00:00:00Z. If you are supplying this, do not supply scheduleCron.",
-          "type": "string",
-          "willReplaceOnChanges": true
```
nyobe, line 1349: I don't think these willReplaceOnChanges should be dropped

pgavlin: Same--these annotations affect doc generation

pgavlin (changes requested):
Could we split the new properties out into their own PR?

IaroslavTitov:
> Could we split the new properties out into their own PR?

Fair, will do

IaroslavTitov:
Updated to only include diff changes

pgavlin:
Do we have diff tests for this?

IaroslavTitov:
> Do we have tests for this?

All of the resources have examples/integ tests that run then up, update, refresh and destroy. So all of these are checked. Also there is no actual functionality change here.

pgavlin:
> Also there is no actual functionality change here.

Yeah that's just difficult to validate by inspection. That said, I think things look okay? Hard to be confident that behavior isn't changing w/o some explicit tests that validate that particular kinds of diffs haven't changed.

IaroslavTitov:
> some explicit tests that validate that particular kinds of diffs haven't changed.

Well, in theory, since none of the tests are breaking, there's no functionality change (that is, if we had full coverage, which we don't). But yeah, anything I can add to make it more convincing? 😅

komalali:
I don't feel confident about this change since we don't have great testing in the service provider, especially at the diff layer. This PR doesn't have any obvious benefit for the user (there is no user facing change if everything works). I'd want to see some thorough tests for the diff methods specfically if we were to check this in.

IaroslavTitov:
Spoke with Komal, I'll close down this PR for now, and potentially bring back this standard diff after breaking changes as part of PSPV1

Issue #500, opened 2025-03-05
Refreshing deleted TtlSchedule fails


IaroslavTitov:
### What happened?

I ran pulumi refresh on a stack that had a TtlSchedule created, and got a 404 error, because it no longer existed. Instead, it should have marked it as non-existant and updated the state.

### Example

pulumi refresh

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #510, opened 2025-03-21
Fixing deleted schedules Read method

IaroslavTitov:
### Summary
- Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/500
- On `GetSchedule` call, return nil for both error and schedule on 404
- This makes schedule resources return empty response and cause the provider to mark object as `delete` on refresh of non-existing schedule (mostly relates to TTL schedules that destroy themselves after executing)

### Testing
- Manually tested by creating a TTL schedule that executed right after, ran `pulumi refresh`

### DIFF
--- a/provider/pkg/pulumiapi/schedules.go
+++ b/provider/pkg/pulumiapi/schedules.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"path"
-	"strings"
 	"time"
 )
 
@@ -105,7 +104,7 @@ func (c *Client) GetSchedule(ctx context.Context, stack StackIdentifier, schedul
 	var scheduleResponse ScheduleResponse
 	_, err := c.do(ctx, http.MethodGet, apiPath, nil, &scheduleResponse)
 	if err != nil {
-		if strings.Contains(err.Error(), "404 API error") {
+		if GetErrorStatusCode(err) == http.StatusNotFound {
 			return nil, nil
 		}
 		return nil, fmt.Errorf("failed to get schedule with scheduleId %s : %w", scheduleID, err)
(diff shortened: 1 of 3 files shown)


provider/pkg/pulumiapi/schedules.go
```diff
@@ -104,6 +105,9 @@ func (c *Client) GetSchedule(ctx context.Context, stack StackIdentifier, schedul
 	var scheduleResponse ScheduleResponse
 	_, err := c.do(ctx, http.MethodGet, apiPath, nil, &scheduleResponse)
 	if err != nil {
+		if strings.Contains(err.Error(), "404 API error") {
```
nyobe, line 108: I think you should check the http error code rather than doing a string match on the error message. c.do will [return an error](https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/pulumiapi/client.go#L94) that carries along the http status code, and there is a [helper to check it](https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/pulumiapi/error.go#L31):

```suggestion
		if GetErrorStatusCode(err) == http. {
```

nyobe (approved):

provider/pkg/pulumiapi/schedules.go
IaroslavTitov: Ah, good idea, completely forgot about this method! Will update to use `GetErrorStatusCode(err) == http.StatusNotFound`

IaroslavTitov:
The tests are breaking due to https://github.com/pulumi/pulumi/issues/19004, since this is not urgent, I will just wait until pu/pu release and re-run the tests

Issue #512, opened 2025-03-26
TTLSchedule Resource ignore `deleteAfterDestroy` property


MitchellGerdisch:
### What happened?

I have a stack that configures a TTL Schedule with the `deleteAfterDestroy` property set to true.
Everything seems to be fine.
But if you go look at the TTL schedule in the UI, you'll see that the "delete after destroy" slider is not enabled.

### Example

The code in this folder can be used to see this:
https://github.com/MitchellGerdisch/pulumi-work/tree/09d119697f3860ecf19de4a3b53c5884c78b7232/ts-deployment-schedules

### Output of `pulumi about`

CLI          
Version      3.158.0
Go Version   go1.24.1
Go Compiler  gc

Plugins
KIND      NAME           VERSION
resource  command        0.11.1
language  nodejs         3.158.0
resource  pulumiservice  0.29.0

Host     
OS       darwin
Version  15.3.2
Arch     x86_64

This project is written in nodejs: executable='/Users/mitch/.nvm/versions/node/v20.10.0/bin/node' version='v20.10.0'

Found no pending operations associated with team-ce/dev

Backend        
Name           pulumi.com

Dependencies:
NAME                   VERSION
@pulumi/command        0.11.1
@pulumi/pulumi         3.158.0
@pulumi/pulumiservice  0.29.0
@types/node            18.19.83
typescript             5.8.2

Pulumi locates its logs in /var/folders/qp/6k0zsrj13rz5ll53hsmlksvw0000gq/T/ by default

### Additional context

Some additional notes:
- once the resource is set via the provider, tryin to change the slider in the UI is ignored and you can't save the updated schedule.
- if you change the slider and the time, then the save schedule button is active, but after saving it, the delete after destroy is still NOT enabled.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #517, opened 2025-04-09
Fix a typo in the TTL API type

komalali:
The typo was preventing the API request from being marshalled correctly, causing the `deleteBeforeDestroy` property to get dropped.

Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/512

### DIFF
--- a/provider/pkg/pulumiapi/schedules.go
+++ b/provider/pkg/pulumiapi/schedules.go
@@ -50,7 +50,7 @@ type CreateDriftScheduleRequest struct {
 
 type CreateTtlScheduleRequest struct {
 	Timestamp          time.Time `json:"timestamp,omitempty"`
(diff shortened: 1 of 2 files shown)


provider/pkg/pulumiapi/schedules.go
```diff
@@ -50,7 +50,7 @@ type CreateDriftScheduleRequest struct {
 
 type CreateTtlScheduleRequest struct {
 	Timestamp          time.Time `json:"timestamp,omitempty"`
-	DeleteAfterDestroy bool      `json:"deleteAfterDestroy ,omitempty"`
+	DeleteAfterDestroy bool      `json:"deleteAfterDestroy,omitempty"`
```
IaroslavTitov, line 53: Why would marshaller allow space at the end of the name ;-;

IaroslavTitov (approved):

PR #525, opened 2025-05-29
teams: add support for ESC env max open duration 

pgavlin:
These changes add support to the TeamEnvironmentPermission resource for
configuring maximum open durations for ESC environments. The max open
duration is represented as a new property on the resource.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ toolchain go1.24.0
 
 require (
 	github.com/google/uuid v1.6.0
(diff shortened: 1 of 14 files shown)


komalali (approved):
lgtm, would be nice to add a test. looks like we don't have an env permission test but maybe we can model one here or turn this one into a generic "team-permissions" test? https://github.com/pulumi/pulumi-pulumiservice/blob/main/examples/yaml-team-stack-permissions/Pulumi.yaml

PR #539, opened 2025-07-24
Adding approval rule resource

IaroslavTitov:
### Notes
- `teamName` approver is untested, b/c pending API changes in `pulumi-service`. Test and update examples before merging!

### Summary
- Fixes https://github.com/pulumi/pulumi-service/issues/29227
- Adding new Approval Rule resource
- Adding examples

### Testing
- Manually tested


examples/ts-approval-rules/index.ts
```diff
@@ -0,0 +1,46 @@
...
+var approvalRule = new service.ApprovalRule("rule-test", {
+  name: "My rule!",
+  description: "My description!",
+  enabled: true,
+  targetActionType: "update",
+  environmentIdentifier: {
+    organization: environment.organization,
+    project: environment.project,
+    name: environment.name,
+  },
```
nyobe, line 29: what do you think of grouping the action type + identifier together as a `target`?
```
target: {
  actionType: "update",
  identifier: {}
}
```

we're also going to end up needing to add different types of identifiers for different entity types as we add them.  what's the best way to model this in a provider?

examples/ts-approval-rules/index.ts
```diff
@@ -0,0 +1,46 @@
...
+`values:
+  myKey1: "myValue1"`
+  )
+})
+
+// Create an approval rule that governs who can approve updates to the environment
+// This rule requires 3 approvals from eligible approvers before any update can proceed
+var approvalRule = new service.ApprovalRule("rule-test", {
+  name: "My rule!",
+  description: "My description!",
```
nyobe, line 22: we're going to drop description so you can leave it out

examples/ts-approval-rules/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  enabled: true,
+  targetActionType: "update",
+  environmentIdentifier: {
+    organization: environment.organization,
+    project: environment.project,
+    name: environment.name,
+  },
+  approvalRuleConfig: {
```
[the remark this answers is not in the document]
IaroslavTitov: Yeah, so I don't believe we'd be able to replicate the tagged union in schema json, so I was going to rely on logic to check method to ensure certain fields are set. We can add `type` to approval rule config and have Check method ensure specific rules are set. Alternatively, can rename this object something more specific like `approvalRuleApproversConfig`, and in the future just add different configs, with Check method verifying only one of them is set. The second option will probably be more user friendly.

examples/ts-approval-rules/index.ts
nyobe: makes sense to me

nyobe (approved):

nyobe (approved):

PR #543, opened 2025-08-01
Update Approval rules apis

seanyeh:
Changes:
* update to use the updated change gate apis
* adds tests
* small fixes to the read api

### DIFF
--- a/provider/pkg/pulumiapi/approvals.go
+++ b/provider/pkg/pulumiapi/approvals.go
@@ -24,10 +24,10 @@ import (
 // Because the API routes require envIdentifiers, I'm making these methods specific, but in the future
 // we might want to generalize them
 type ApprovalRuleClient interface {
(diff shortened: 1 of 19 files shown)


provider/pkg/pulumiapi/approvals.go
```diff
@@ -64,16 +64,21 @@ type EligibleApprover struct {
 
 type EligibleApproverOutput struct {
 	EligibilityType ApprovalRuleEligibilityType `json:"eligibilityType"`
-	TeamName        string                      `json:"teamName,omitempty"`
-	GithubLogin     string                      `json:"githubLogin,omitempty"`
+	TeamName        string                      `json:"name,omitempty"`
+	User            UserInfo                    `json:"user,omitempty"`
```
seanyeh, line 68: this fixes an issue with read because we didn't have the correct schema

PR #583, opened 2025-10-28
docs: adding default values to boolean inputs

### LINKED ISSUE #498: Docs: What is the default value for `pulumiservice.Stack.forceDestroy`?
Image

Yeah, I would probably assume it's `false`, but we should just write that down so that people don't have to run a live experiment to verify.

rshade:
Properties Documented ✅
1. Stack.forceDestroy (default: false)
 - Updated both properties and inputProperties sections
 - Enhanced description: "Defaults to false, which means the stack will not be deleted if it still contains resources."
2. AgentPool.forceDestroy (default: false)
 - Updated both properties and inputProperties sections
 - Enhanced description: "Defaults to false, which means the agent pool will not be deleted if stacks are still configured to use it."
3. OrgAccessToken.admin (default: false)
 - Updated both properties and inputProperties sections
 - Enhanced description: "Optional. True if this is an admin token. Defaults to false."
4. TtlSchedule.deleteAfterDestroy (default: false)
 - Updated both properties and inputProperties sections
 - Enhanced description: "Defaults to false, which means only the resources are destroyed but the stack itself remains."

Fixes #498

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,21 @@
 
 ### Improvements
 
+- Added comprehensive error handling and edge case tests for 6 high-priority resources [#578](https://github.com/pulumi/pulumi-pulumiservice/issues/578):
+  - AccessToken: 19 new test functions covering API errors, validation, and edge cases
+  - Stack: 26 new test functions covering create/delete errors, forceDestroy behavior, and special characters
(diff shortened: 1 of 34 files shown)


iwahbe (commented):
This PR seems to have combined default propagation fixes for `forceDestroy` (agent_pool & stack) and docs changes.

I think both are good, but the semantic changes need a test before merging.

PR #591, opened 2025-10-29
Extract the schema with `pulumi package get-schema`

iwahbe:
This breaks the cyclic build between `schema.json` and
`bin/pulumi-resource-pulumiservice`. This ensures that `schema.json` contains the same
result as a user running `pulumi package get-schema`.

This change is an extremely useful prerequisite to generating part of `schema.json`, as
would be involved in a partial conversion to a `pulumi-go-provider` based approach.

There should be no semantic changes in the contents of `schema.json`. `manual-schema.json`
is currently an exact copy of the old `schema.json`.


rshade (approved):
LGTM

Issue #288, opened 2024-05-23
MIgrate provider to updated go-provider-sdk 


IaroslavTitov:
## Issue details

Rebase https://github.com/pulumi/pulumi-pulumiservice/pull/258, merge it in to migrate PSP, Ian did most of the work here.

### Affected area/feature

Provider

Issue #587, opened 2025-10-28
Bug: StandardDiff detects false changes when properties are identical


rshade:
## Problem

The `StandardDiff()` utility function in `provider/pkg/util/diff.go` reports `DIFF_SOME` (changes detected) even when the old and new property values are completely identical.

## Impact

Resources using `StandardDiff()` report unnecessary updates, potentially causing:
- Unnecessary resource refreshes
- Confusing user experience ("no changes" shows as "changes detected")
- Possible unnecessary API calls

## Root Cause

Likely in `provider/pkg/util/diff.go` lines 9-24:
- The unmarshaling of `OldInputs` uses `SkipNulls: true`
- The unmarshaling of `News` uses `SkipNulls: true`
- The diff comparison may be incorrectly handling empty objects or default values

Possible causes:
1. Empty object handling (`{}` vs missing)
2. Null/undefined value differences
3. Array/slice comparison issues
4. Type conversion differences

## Expected Behavior

When `req.Olds` and `req.News` contain identical property values, the response should be:
- `Changes: pulumirpc.DiffResponse_DIFF_NONE` (value = 1)
- `DetailedDiff: {}` (empty map)

## Actual Behavior

Even with identical inputs, the response shows:
- `Changes: pulumirpc.DiffResponse_DIFF_SOME` (value = 2)
- `DetailedDiff: {...}` (contains spurious diffs)

## Test Case

`TestApprovalRule_Diff_NoChanges` in `provider/pkg/resources/approvalRule_test.go:870-901`:
- Sets both Olds and News to the same structpb.Struct
- Expected: `Changes == DIFF_NONE` (1)
- Actual: `Changes == DIFF_SOME` (2)

## Steps to Reproduce

```bash
cd provider/pkg
go test -v ./resources/ -run TestApprovalRule_Diff_NoChanges
```

## Suggested Investigation

1. Add debug logging in StandardDiff to see what `diffs` contains
2. Check if `olds.Diff(news)` correctly handles empty nested objects
3. Verify MarshalOptions consistency
4. Test with simple flat properties vs nested objects

## Files Affected

- `provider/pkg/util/diff.go` (lines 8-49)
- All resources using `StandardDiff()`

## Related

Discovered by unit tests added in #574

Issue #586, opened 2025-10-28
Bug: StandardDiff doesn't populate Replaces array from DetailedDiff


rshade:
## Problem

The `StandardDiff()` utility function in `provider/pkg/util/diff.go` populates the `DetailedDiff` map with replacement properties but does not extract those properties into the `Replaces[]` array.

This causes the `DiffResponse.Replaces` field to always be empty/nil, even when properties are marked as replacements.

## Impact

Resources using `StandardDiff()` (like ApprovalRule, TemplateSource, OidcIssuer, etc.) don't correctly report which property changes require resource replacement.

## Root Cause

In `provider/pkg/util/diff.go` lines 26-38:
- Properties in `replaceProperties` map are marked with `.AsReplace()` in DetailedDiff
- But these properties are never extracted back into the `Replaces[]` array

## Expected Behavior

When a property is marked as a replacement in `replaceProperties`, it should appear in both:
1. `DetailedDiff` (currently working) 
2. `Replaces[]` array (currently missing)

## Actual Behavior

`Replaces[]` is always empty, regardless of replacement properties.

## Test Case

`TestApprovalRule_Diff_EnvironmentChange` in `provider/pkg/resources/approvalRule_test.go:770-819` expects `Replaces` to contain `"environmentIdentifier.organization"` but receives empty array.

## Proposed Fix

After line 38 in `diff.go`, extract replacement properties:

```go
replaces := []string{}
for k, v := range dd {
    if v.Kind.IsReplace() {
        replaces = append(replaces, k)
    }
}
```

Then add to return statement:
```go
return &pulumirpc.DiffResponse{
    Changes:         changes,
    Replaces:        replaces,  // Add this
    DetailedDiff:    detailedDiffs,
    HasDetailedDiff: true,
}, nil
```

## Files Affected

- `provider/pkg/util/diff.go` (lines 8-49)
- All resources using `StandardDiff()`: ApprovalRule, TemplateSource, OidcIssuer, EnvironmentVersionTag, DeploymentSchedule

## Related

Discovered by unit tests added in #574

PR #593, opened 2025-10-29
Convert `pulumiservice:index:Team` to be `infer` based

iwahbe:
#### `pulumiservice:index:Team`

This PR converts `pulumiservice:index:Team` into a [pulumi-go-provider](https://github.com/pulumi/pulumi-go-provider) [infer](https://github.com/pulumi/pulumi-go-provider/tree/main/infer#infer) based resource. Visible behavior should not change for the provider.

#### `infer`

This PR spearheads how new resources built with `infer` can be added to the provider. Using `infer` would prevent bugs like https://github.com/pulumi/pulumi-pulumiservice/issues/594, https://github.com/pulumi/pulumi-pulumiservice/issues/587 and https://github.com/pulumi/pulumi-pulumiservice/issues/586, in addition to lessening the work required to add or update new resources: default implementations of Diff and an automatically generated schema entry.

#### Migrating old resources & functions

There is no need to migrate old resources or functions. They will continue to work as they have in the past. New resources should be added using `infer`.

---

Needed before merging:
- [x] Unskip the unit test, adjusting it to work with the new client injection mechanism.
  - This will require allowing the client to be mocked, then injected.
- [x] Audit how we treat `*string` in inputs to make sure we are not introducing an accidental nil pointer panic with this change.
- [x] Decide and document the DiffConfig semantics. I'm thinking something like in-place update for PULUMI_ACCESS_TOKEN and probably cascading replace for PULUMI_BACKEND_URL.
- [x] Target a release version of `pulumi-go-provider`.

Fixes https://github.com/pulumi/pulumi-pulumiservice/issues/288


sdk/python/pulumi_pulumiservice/team.py
```diff
@@ -212,6 +212,8 @@ def _internal_init(__self__,
             if team_type is None and not opts.urn:
                 raise TypeError("Missing required property 'team_type'")
             __props__.__dict__["team_type"] = team_type
+        replace_on_changes = pulumi.ResourceOptions(replace_on_changes=["name", "organizationName", "teamType"])
```
jdavredbeard: aha! provider struct tag at work

jdavredbeard: no other changes to the sdk, nice

PR #604, opened 2025-11-13
Add review guidelines for sdk.instructions.md

rshade:
Clarify review instructions for SDK files.


iwahbe (commented):
Is there some standard for `.github/sdk.instructions.md`? If so, can you link it?

It's not obvious why this is the best way to describe the `sdk` folder otherwise.

rshade:
> Is there some standard for `.github/sdk.instructions.md`? If so, can you link it?
> 
> It's not obvious why this is the best way to describe the `sdk` folder otherwise.

Not really, I wish we had one. We really just need copilot to ignore the whole thing unless the schema changes, but maybe people will just know that the unclean checkout means run `make build_sdks`, and we can just do #4 on here: https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#creating-path-specific-custom-instructions-2, what do you think

PR #629, opened 2025-12-24
Add support for InsightsAccount resources in policy groups. Fix diff …

fallimic:
- Adds a new "accounts" field to policy groups to support linking InsightsAccount resources
- Updates PolicyGroup to use batch api for updates
- Fixes various issues with refresh/diff false positives

### DIFF
--- a/provider/pkg/resources/policy_group_test.go
+++ b/provider/pkg/resources/policy_group_test.go
@@ -16,6 +16,7 @@ package resources
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


jdavredbeard:
looks good, what makes the batch update api more efficient?

fallimic:
> looks good, what makes the batch update api more efficient?

Just the fact that we can make a single API call for all updates instead of making an API call for each update. It also helps keep the update atomic so we don't have partial failures where some of the updates could succeed and others fail.

Issue #552, opened 2025-10-03
[RBAC] Role resource


jdavredbeard:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
Roles should be modeled as resources

PR #664, opened 2026-02-09
Add roleID to OIDC Issuer policies

altanozlu:
Pulumi IAM supports custom roles for OIDC's organizational policies, which can be created through the UI, but the provider doesn't support them yet. This PR aims to resolve that. 
Since #552 isn't completed yet, roles must already be created outside of this provider.


altanozlu:
@iwahbe can you check again ?

iwahbe:
The PR looks reasonable to me.

iwahbe (approved):

PR #668, opened 2026-02-24
[deployments] add executorRootPath setting

rgharris:
Closes https://github.com/pulumi/pulumi-service/issues/38270

Note: API support for this is going out in the next prod service deployment


rshade (approved):
LGTM -> Very cool addition

provider/pkg/pulumiapi/deployment_settings.go
```diff
@@ -26,13 +26,18 @@ type DeploymentSettingsClient interface {
...
+	OperationContext *OperationContext    `json:"operationContext,omitempty"`
+	GitHub           *GitHubConfiguration `json:"gitHub,omitempty"`
+	SourceContext    *SourceContext       `json:"sourceContext,omitempty"`
+	ExecutorContext  *ExecutorContext     `json:"executorContext,omitempty"`
+	AgentPoolID      string               `json:"agentPoolId,omitempty"`
+	Source           *string              `json:"source,omitempty"`
+	CacheOptions     *CacheOptions        `json:"cacheOptions,omitempty"`
+}
+
+type ExecutorContext struct {
```
seanyeh: asking as i'm not very familiar - I see that the apitype version has a WorkingDirectory. how does that differ from ExecutorRootPath? it does seem weird we have two different `ExecutorContext` structs

seanyeh (approved):

PR #671, opened 2026-03-02
[VCS] Add VCS support to DeploymentSettings

lukebward:
- Added vcs property to DeploymentSettings for polymorphic VCS configuration (Azure DevOps and GitHub), which is the correct way to reference an existing integration
  - Added YAML example for DeploymentSettings VCS configuration
  - Added utility functions for handling secret/non-secret string values

  Test plan

  - Provider builds successfully
  - Provider unit tests pass (go test -short ./...)
  - Linting passes
  - SDKs regenerated without AzureDevOpsIntegration
  - CI runs full test suite
  - TestYamlDeploymentSettingsVcsExample passes against a configured ADO environment (manual, requires OAuth setup)

### DIFF
--- a/provider/pkg/provider/manual-schema.json
+++ b/provider/pkg/provider/manual-schema.json
@@ -378,6 +378,10 @@
             {
               "value": "github",
               "name": "GitHub"
+            },
+            {
+              "value": "gitlab",
+              "name": "GitLab"
             }
           ]
         },


provider/pkg/pulumiapi/deployment_settings.go
fallimic: Is there a way to mark this as deprecated in the manual-schema.json?

examples/examples_yaml_test.go
```diff
@@ -474,6 +474,10 @@ func TestYamlInsightsAccountExample(t *testing.T) {
 	runPulumiTest(t, test)
 }
 
+func TestYamlDeploymentSettingsVcsExample(t *testing.T) {
+	t.Skip("requires an existing Azure DevOps integration; run manually against a configured environment")
```
fallimic: Can we set up an integration on the test account that we use in ci? I think it's the staging moolumi org?

provider/pkg/provider/manual-schema.json
```diff
@@ -364,6 +364,57 @@
       },
       "type": "object"
     },
+    "pulumiservice:index:DeploymentSettingsVcs": {
+      "description": "VCS settings for the deployment, supporting multiple VCS providers.",
+      "properties": {
+        "provider": {
+          "type": "string",
+          "description": "The VCS provider type.",
+          "enum": [
```
fallimic: Should we add gitlab as well given the imminent support?

PR #673, opened 2026-03-06
Fix Environment resource spurious replace on upgrade from pre-0.25.0

borisschlosser:
## Summary

Fixes a bug where upgrading the Pulumi Service Provider from pre-0.25.0 to 0.25.0+ causes **all existing Environment resources to be spuriously replaced** (deleted and recreated).

### Root Cause

Version 0.25.0 introduced the `project` input property to the Environment resource (defaulting to `"default"`). When performing a diff, old state from pre-0.25.0 lacks the `project` field entirely. The Diff method compares `nil` vs `"default"`, producing a diff on `project` — which is a replace-triggering property — and forces a delete-and-recreate of every Environment resource.

### Fix

**Backfill missing `project` in old inputs**: When old inputs (from pre-0.25.0 state) don't have a `project` field, we inject `project: "default"` before computing the diff. This eliminates the false diff that was causing the spurious replace.

### Test coverage

Added `TestEnvironmentDiff` with two sub-tests:
- **No spurious replace on upgrade**: Verifies that old inputs without `project` + new inputs with `project: "default"` produces `DIFF_NONE`
- **Replace when project actually changes**: Verifies that a genuine project change (`proj-a` → `proj-b`) still triggers a replace

## Test plan

- [x] `TestEnvironmentDiff/No_spurious_replace_when_upgrading_from_pre-0.25.0_state_without_project` — passes
- [x] `TestEnvironmentDiff/Replace_when_project_actually_changes` — passes
- [x] All existing `TestEnvironment` and `TestEnvironmentCheck` tests — pass
- [x] Manual verification: with old state lacking `project`, `pulumi preview` should show no replace

### DIFF
--- a/provider/pkg/resources/environment.go
+++ b/provider/pkg/resources/environment.go
@@ -103,7 +103,7 @@ func getBytesFromAsset(asset *asset.Asset) ([]byte, error) {
 func (st *PulumiServiceEnvironmentResource) Diff(req *pulumirpc.DiffRequest) (*pulumirpc.DiffResponse, error) {
 	olds, err := plugin.UnmarshalProperties(
 		req.GetOldInputs(),
-		plugin.MarshalOptions{KeepUnknowns: false, SkipNulls: true},
+		plugin.MarshalOptions{KeepUnknowns: true, SkipNulls: true},
 	)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


borisschlosser:
> I'd stick with the first change from the change list. That should be the minimal edit necessary to fix the problem.

Thanks @iwahbe reverted the other points, keeping them in the PR description as possible follow ups.

iwahbe (approved):

PR #731, opened 2026-04-07
Add TwelveHours (12h) scan schedule for Insights accounts

lblackstone:
## Summary

- Adds `TwelveHours` / `"12h"` enum value to `ScanSchedule` for twice-daily Insights account scanning
- Regenerates schema and all language SDKs (Go, TypeScript, Python, .NET, Java)
- Companion to pulumi/pulumi-service#40828 which adds the same enum value on the service side

## Test plan

- [x] Provider tests pass (`go test ./provider/pkg/resources/ -run TestInsightsAccount`)
- [x] CI passes
- [ ] Verify `12h` schedule can be set via Pulumi program after both PRs merge

### DIFF
--- a/provider/cmd/pulumi-resource-pulumiservice/schema.json
+++ b/provider/cmd/pulumi-resource-pulumiservice/schema.json
@@ -12,9 +12,6 @@
   "repository": "https://github.com/pulumi/pulumi-pulumiservice",
   "publisher": "Pulumi",
   "namespace": "pulumi",
-  "meta": {
-    "moduleFormat": "(.*)"
-  },
   "language": {
     "csharp": {
       "namespaces": {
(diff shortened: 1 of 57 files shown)


provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -12,9 +12,6 @@
   "repository": "https://github.com/pulumi/pulumi-pulumiservice",
   "publisher": "Pulumi",
```
fallimic: Curious why there are so many unrelated changes here. I understand this is generated, but it's weird that it is so non-deterministic. It's most likely related to pulumi cli version.

fallimic (approved):

PR #748, opened 2026-04-19
Add StackTags resource for managing multiple stack tags

### LINKED ISSUE #61: Add StackTags resource
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

We should add a `StackTags` resource in addition to our `StackTag` resource. Right now, we have a "StackTag" resource. This works well for adding a single stack tag, however for convenience it would be nice to define a map of stack tags in a single resource. A map would be particularly nice for YAML support, where otherwise we'd need to repeat the same keys over and over again.

```yaml
  stackTags:
    type: pulumiservice:StackTags
    properties:
      tags:
        tag1: value1
        tag2: value2
        tag3: value3
```

AnuiDolphin:
## Summary

Adds a new `pulumiservice:index:StackTags` resource that manages a set of stack tags as a single resource, with a `tags` map input — closes #61.

```yaml
myTags:
  type: pulumiservice:StackTags
  properties:
    organization: my-org
    project: my-project
    stack: my-stack
    tags:
      environment: production
      team: platform
      cost-center: engineering
```

### Behavior
- Full CRUD. Changes to `organization` / `project` / `stack` replace; `willReplaceOnChanges` is declared in the schema so the planner surfaces replacement without running the provider.
- Pulumi Cloud tags are immutable per key, so a value change is implemented as delete + create.
- `Read` only returns tags that were previously in state, so this resource never clobbers tags managed elsewhere (CLI, pulumibot, a singular `StackTag` resource). Import via `{org}/{project}/{stack}/tags` adopts every tag on the stack — the schema description guides users to declare `tags` explicitly after import.
- On partial failure in Create or Update, the returned error carries an `ErrorResourceInitFailed` detail listing the tags that did succeed, so Pulumi records them in state and the next `up` doesn't recreate tags that already exist. A unit test exercises this path.

### Attribution

This PR leverages the work in @rshade's PR #548. The resource implementation, API client addition, schema shape, SDK structure, unit test harness, and YAML example originate there. This version rebases onto current main (after the `util.FromProperties`/`ToProperties` signature change and the `manual-schema.json` split in #591), adds the partial-failure handling, declares `willReplaceOnChanges` in the schema, expands the schema description with import semantics, and addresses the Copilot review comments on #548.

## Test plan

- [x] `go test ./provider/...` — all provider tests pass
- [x] `go test -run TestStackTagsPlural -v ./provider/pkg/resources/...` — all 9 subtests pass, including the new `CreatePartialFailure` case
- [x] `make schema` regenerates `schema.json` cleanly from `manual-schema.json`
- [x] `make generate_go generate_nodejs generate_python generate_dotnet` — SDKs regenerate matching the committed output
- [x] `make lint` — no new issues on changed files (7 pre-existing gosec findings on webhook/config/etc. remain)
- [ ] YAML integration test (`TestYamlStackTagsPluralExample`) — exercised by CI


provider/pkg/resources/stack_tags_plural.go
```diff
@@ -0,0 +1,347 @@
...
+		KeepOutputValues: true,
+	})
+}
+
+func (st *PulumiServiceStackTagsResource) ToPulumiServiceStackTagsInput(
+	inputMap resource.PropertyMap,
+) PulumiServiceStackTagsInput {
+	input := PulumiServiceStackTagsInput{}
+	_ = util.FromPropertyMap(inputMap, &input)
+	return input
```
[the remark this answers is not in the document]
AnuiDolphin: Fixed: `ToPulumiServiceStackTagsInput` now returns `(PulumiServiceStackTagsInput, error)` and the one test caller was updated to handle the error.

provider/pkg/resources/stack_tags_plural.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+
+	// Parse current inputs to know which tags we manage.
+	var currentInput PulumiServiceStackTagsInput
+	if req.Inputs != nil {
+		inputMap, err := plugin.UnmarshalProperties(req.Inputs, plugin.MarshalOptions{KeepUnknowns: false, SkipNulls: true})
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal inputs: %w", err)
+		}
+		_ = util.FromPropertyMap(inputMap, &currentInput)
```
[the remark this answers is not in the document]
AnuiDolphin: Fixed: `Read` now returns an error if decoding `req.Inputs` fails. This prevents the resource from silently falling into the adopt-all-tags path after a decode failure.

provider/pkg/util/property_map.go
```diff
@@ -137,6 +151,33 @@ func set(v reflect.Value, value interface{}) error {
...
+				v.Set(reflect.MakeMap(v.Type()))
+			}
+			// Populate the map
+			for k, val := range mapValue {
+				strVal, ok := val.(string)
+				if !ok {
+					return fmt.Errorf("expected string value in map, got %T", val)
+				}
+				v.SetMapIndex(reflect.ValueOf(k), reflect.ValueOf(strVal))
+			}
```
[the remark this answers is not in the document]
AnuiDolphin: Fixed: `set` now always starts from a fresh map (`reflect.MakeMapWithSize`) rather than writing into an existing one, so stale keys from a prior decode into the same struct can't linger. Added a regression test (`Replaces map fields rather than merging into existing values`) in `property_map_test.go`.

examples/examples_utils_test.go
```diff
@@ -27,3 +27,10 @@ func getOrgName() string {
 	}
 	return "service-provider-test-org"
 }
+
+func getTestOrg() string {
+	if org := os.Getenv("PULUMI_TEST_OWNER"); org != "" {
+		return org
+	}
+	return "service-provider-test-org"
```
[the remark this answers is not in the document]
AnuiDolphin: Fixed: removed `getTestOrg` and updated the one call site in `examples_yaml_test.go` to use the existing `getOrgName()`.

AnuiDolphin:
Addressed all four Copilot review comments in 8ea9077 (force-pushed):

| # | Comment | Fix |
|---|---------|-----|
| 1 | `ToPulumiServiceStackTagsInput` swallowed decode errors with `_ = util.FromPropertyMap(...)` | Signature is now `(PulumiServiceStackTagsInput, error)`; the one test caller was updated to check the error. |
| 2 | `Read` silently fell into the adopt-all-tags path on decode failure | `Read` now returns the decode error instead of ignoring it, so a corrupt state can't cause the resource to take ownership of tags it never managed. |
| 3 | `set` in `property_map.go` left stale keys when decoding a map into an already-populated struct | Always allocate a fresh map via `reflect.MakeMapWithSize` before populating. Added a regression test `Replaces map fields rather than merging into existing values` in `property_map_test.go`. |
| 4 | `getTestOrg` duplicated `getOrgName` in `examples_utils_test.go` | Removed `getTestOrg`; updated the single call site in `examples_yaml_test.go` to use `getOrgName()`. |

Each review thread has an individual reply with the same note.

rshade (approved):
LGTM

AnuiDolphin:
Also dropped the two new indirect dependencies (`github.com/joho/godotenv` and `github.com/json-iterator/go`) that had been added to the root `go.mod`. `go mod why` confirmed no code in the module imports either of them — they were cruft from a `go mod tidy` run with unrelated tooling on the path. `go mod tidy` in a clean environment removes them, and the PR now has a zero diff in `go.mod`/`go.sum` against `main`.

examples/examples_utils_test.go
```diff
@@ -27,3 +27,4 @@ func getOrgName() string {
 	}
 	return "service-provider-test-org"
 }
+
```
davidemassarenti-optio3: Only change in the file. Revert?

provider/pkg/util/property_map.go
```diff
@@ -102,6 +102,16 @@ func get(v reflect.Value) interface{} {
 			return get(v.Elem())
 		}
 		return nil
+	case reflect.Map:
+		// Handle map[string]string
+		if v.Type().Key().Kind() == reflect.String && v.Type().Elem().Kind() == reflect.String {
+			result := make(map[string]interface{})
```
davidemassarenti-optio3: Not map[string]string? We are checking for that in v

examples/examples_utils_test.go
AnuiDolphin: Good catch — that was an accidental trailing-newline leftover from removing `getTestOrg`. Reverted so `examples_utils_test.go` is no longer in the diff.

provider/pkg/util/property_map.go
AnuiDolphin: Fixed — `get` now returns `map[string]string` directly (pre-sized via `make(..., v.Len())`) instead of boxing through `map[string]interface{}`. Tests still pass.

provider/pkg/util/property_map.go
```diff
@@ -137,6 +151,31 @@ func set(v reflect.Value, value interface{}) error {
 		v.Set(reflect.New(v.Type().Elem()))
 		// call set again, but with deref'd value. note that this will recurse down for ptr to ptr's (and so on)
 		return set(v.Elem(), value)
+	case reflect.Map:
+		// Handle map[string]string
+		if v.Type().Key().Kind() == reflect.String && v.Type().Elem().Kind() == reflect.String {
+			var mapValue map[string]interface{}
+
+			// Check if value is a PropertyMap and convert it
+			if pm, ok := value.(resource.PropertyMap); ok {
```
davidemassarenti-optio3: Can this ever be true? PropertyMap's values are structs, not strings.

provider/pkg/util/property_map.go
```diff
@@ -137,6 +151,31 @@ func set(v reflect.Value, value interface{}) error {
 		v.Set(reflect.New(v.Type().Elem()))
 		// call set again, but with deref'd value. note that this will recurse down for ptr to ptr's (and so on)
 		return set(v.Elem(), value)
+	case reflect.Map:
+		// Handle map[string]string
+		if v.Type().Key().Kind() == reflect.String && v.Type().Elem().Kind() == reflect.String {
+			var mapValue map[string]interface{}
```
davidemassarenti-optio3: var mapValue map[string]string

provider/pkg/util/property_map.go
AnuiDolphin: You're right — that branch was unreachable. `set` is only called via `FromPropertyMap` → `mapVal.V`, and for map-valued properties `PropertyValue.V` is always `resource.PropertyMap` (`map[resource.PropertyKey]PropertyValue`), never a raw `map[string]interface{}`. Removed the dead branch; `set` now fails fast with `expected resource.PropertyMap, got %T` if anything else comes in.

provider/pkg/util/property_map.go
AnuiDolphin: Good catch — the intermediate `map[string]interface{}` (and the `pm.Mappable()` call that produced it) was gratuitous boxing since we already know the target is `map[string]string`. Rewrote the loop to iterate the `resource.PropertyMap` directly, assert `IsString()` per value, and call `SetMapIndex` with the concrete `StringValue()`. No `any` on the hot path.

provider/pkg/resources/stack_tags_plural.go
```diff
@@ -0,0 +1,354 @@
...
+			state.Tags = currentTags
+			return nil, partialErrorStackTags(id, fmt.Errorf("failed to delete tag %q: %w", tagName, err), state, newInput)
+		}
+		delete(currentTags, tagName)
+	}
+
+	createNames := sortedKeys(tagsToCreate)
+	for _, name := range createNames {
+		value := tagsToCreate[name]
+		err := st.Client.CreateTag(ctx, stackName, pulumiapi.StackTag{Name: name, Value: value})
```
davidemassarenti-optio3: Shouldn't this be CreateStackTag?

AnuiDolphin: The asymmetric naming (`CreateTag` vs. `GetStackTag` / `DeleteStackTag`) was pre-existing in the API client. Renamed `CreateTag` → `CreateStackTag` on both the `StackTagClient` interface and the `*Client` receiver. Updated all callers: the singular `StackTag` resource, both call sites in `stack_tags_plural.go`, and the two test call sites in `stack_tags_test.go`. All three tag operations on the client now follow the same `*StackTag` convention.

davidemassarenti-optio3 (approved):

AnuiDolphin (commented):
Review: 9 inline comments — 3 important, 5 suggestions, 1 nitpick. The important items are all test-coverage gaps on the bug classes this PR is specifically designed to prevent (partial-failure on Update, import adopt-all, decode-error path). Everything else is polish.

provider/pkg/resources/stack_tags_plural_test.go
```diff
@@ -0,0 +1,525 @@
...
+		for _, tag := range createdTags {
+			tagMap[tag.Name] = tag.Value
+		}
+		assert.Equal(t, "value1", tagMap["tag1"])
+		assert.Equal(t, "value2", tagMap["tag2"])
+		assert.Equal(t, "value3", tagMap["tag3"])
+	})
+}
+
+func TestStackTagsPluralCreatePartialFailure(t *testing.T) {
```
AnuiDolphin, line 95: **[Important]** Missing symmetric test for `Update` partial-failure. `TestStackTagsPluralCreatePartialFailure` covers Create, but the Update path (delete-then-create mid-batch failure) has no coverage. Update's bookkeeping is actually more complex — `currentTags` is mutated across both the delete and create loops, so the failure state depends on which loop trips. Please add a test that simulates a 500 on the N'th delete or create and asserts `ErrorResourceInitFailed.Properties.Tags` reflects the actual live set at the point of failure.

provider/pkg/resources/stack_tags_plural.go
```diff
@@ -0,0 +1,354 @@
...
+		for tagName := range currentInput.Tags {
+			if value, exists := allTags[tagName]; exists {
+				managedTags[tagName] = value
+			}
+		}
+	} else {
+		// No previous state (import scenario): adopt every tag currently on the stack.
+		// The user is expected to declare the `tags` map explicitly after import so
+		// subsequent updates match their intent — see the schema description.
+		managedTags = allTags
```
AnuiDolphin: **[Important]** No test for the import/adopt-all branch. `Read` with no prior inputs silently takes ownership of every tag on the stack — a high-impact behavior that's only documented in the schema description. Please add a test that calls `Read` with `req.Inputs == nil` against a mock returning three tags and asserts all three end up in the returned state.

provider/pkg/resources/stack_tags_plural.go
```diff
@@ -0,0 +1,354 @@
...
+	if req.Inputs != nil {
+		inputMap, err := plugin.UnmarshalProperties(req.Inputs, plugin.MarshalOptions{KeepUnknowns: false, SkipNulls: true})
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal inputs: %w", err)
+		}
+		if err := util.FromPropertyMap(inputMap, &currentInput); err != nil {
+			// If we can't decode the previous inputs, don't silently fall into
+			// the import/adopt-all-tags path — that would let us take ownership
+			// of (and later delete) tags we never managed.
+			return nil, fmt.Errorf("failed to decode stored inputs for %q: %w", req.Id, err)
```
AnuiDolphin, line 171: **[Important]** No test for the decode-error path. The `return nil, fmt.Errorf("failed to decode stored inputs...")` branch was added in response to earlier review but isn't exercised. Please add a test that constructs a malformed `req.Inputs` (e.g. wrong type for `tags`) and asserts `Read` returns the decode error rather than falling into adopt-all. Without this, a regression reintroducing the `_ = util.FromPropertyMap(...)` pattern would pass CI.

provider/pkg/pulumiapi/stack_tags.go
```diff
@@ -82,3 +82,17 @@ func (c *Client) DeleteStackTag(ctx context.Context, stackName StackIdentifier,
 	}
 	return nil
 }
+
+// GetStackTags retrieves all tags for a stack
+func (c *Client) GetStackTags(ctx context.Context, stackName StackIdentifier) (map[string]string, error) {
```
AnuiDolphin, line 85: **[Suggestion]** `GetStackTag` and `GetStackTags` issue the identical HTTP request and parse the same `stack` struct — the singular one just looks up one key afterward. Refactor `GetStackTag` to delegate: `tags, err := c.GetStackTags(ctx, stackName); ...`. Signature stays the same, no caller changes, no observable behavior change (no test asserts on the wrapped-error prefix).

provider/pkg/resources/stack_tags_plural.go
```diff
@@ -0,0 +1,354 @@
...
+	return util.ToPropertyMap(*i)
+}
+
+func (i *PulumiServiceStackTagsInput) ToRPC() (*structpb.Struct, error) {
+	return plugin.MarshalProperties(i.ToPropertyMap(), plugin.MarshalOptions{
+		KeepOutputValues: true,
+	})
+}
+
+func (st *PulumiServiceStackTagsResource) ToPulumiServiceStackTagsInput(
```
AnuiDolphin: **[Suggestion]** `ToPulumiServiceStackTagsInput` is only called from a single PropertyMap round-trip test — Create/Read/Update all go through `util.FromProperties` directly. Either drop the helper or route the other methods through it for consistency.

provider/pkg/resources/stack_tags_plural.go
AnuiDolphin: **[Suggestion]** `managedTags = allTags` aliases the map returned by `GetStackTags`. No caller currently mutates it, but if one ever does, the mutation would leak into the API-client-owned map. Cheap defensive copy via `maps.Clone(allTags)` removes the footgun.

provider/pkg/util/property_map.go
```diff
@@ -137,6 +151,24 @@ func set(v reflect.Value, value interface{}) error {
 		v.Set(reflect.New(v.Type().Elem()))
 		// call set again, but with deref'd value. note that this will recurse down for ptr to ptr's (and so on)
 		return set(v.Elem(), value)
+	case reflect.Map:
```
AnuiDolphin, line 159: **[Suggestion]** `get` and `set` silently handle only `map[string]string` — any other map type returns `nil` / no-op without an error. A future struct with `map[string]int` would decode to an empty map and nobody would notice. Either return an error for unsupported key/value kinds, or document the restriction on the exported `FromPropertyMap` / `ToPropertyMap` functions.

provider/pkg/provider/manual-schema.json
```diff
@@ -1410,6 +1410,66 @@
         "stack"
       ]
     },
+    "pulumiservice:index:StackTags": {
+      "description": "`StackTags` manages a set of stack tags as a single resource. Tags are supplied as a map and are created, updated, and deleted together.\n\nThis is especially convenient for YAML programs, where a single resource can declare many tags rather than needing one `StackTag` resource per key.\n\nThe resource tracks the set of tags it manages in state; tags added to the stack outside this resource (e.g. via the CLI, pulumibot, or a singular `StackTag` resource) are left alone. Because the Pulumi Cloud tag API is immutable per key, changing a tag value is implemented as delete-and-recreate.\n\n### Import\n\nImporting with an ID of `{organization}/{project}/{stack}/tags` adopts **every** tag currently on the stack into this resource's managed set. After import, explicitly declare the `tags` map in your program so subsequent updates match your intent.\n\n```sh\n$ pulumi import pulumiservice:index:StackTags mytags my-org/my-project/my-stack/tags\n```\n",
```
AnuiDolphin: **[Suggestion]** The description embeds a fenced `pulumi import` example plus several paragraphs. It renders fine on the registry but is heavy in CLI `pulumi docs` output. Consider trimming to a short prose summary plus a link to the registry page for the full import example.

provider/pkg/resources/stack_tags_plural.go
```diff
@@ -0,0 +1,354 @@
...
+
+	// Create tags in sorted order so partial failures are deterministic.
+	created := map[string]string{}
+	tagNames := sortedKeys(input.Tags)
+	for _, name := range tagNames {
+		err := st.Client.CreateStackTag(ctx, stackName, pulumiapi.StackTag{Name: name, Value: input.Tags[name]})
+		if err != nil {
+			// Record the subset of tags that were successfully created so Pulumi
+			// can track them in state and the next `up` resumes from the right place.
+			partial := input
```
AnuiDolphin, line 128: **[Nitpick]** `partial := input` followed by `partial.Tags = created` relies on the struct-value-with-map-pointer semantics to swap just the `Tags` field without leaking. It's correct, but a one-line comment (`// copy the value struct but swap Tags to the subset that succeeded`) would save a future reader the mental trace.

provider/pkg/resources/stack_tags_plural_test.go
AnuiDolphin: Addressed in 5890487. Added `TestStackTagsPluralUpdatePartialFailure` with two subtests — one that fails on the N'th delete, one that fails on the N'th create — and asserts `ErrorResourceInitFailed.Properties.Tags` reflects the actual live set at the point of failure in each case.

provider/pkg/resources/stack_tags_plural.go
AnuiDolphin: Addressed in 5890487. Added `TestStackTagsPluralRead / Import adopts every tag on the stack when there is no prior state` — calls `Read` with `req.Inputs == nil` against a mock returning three tags and asserts all three land in the returned state.

provider/pkg/resources/stack_tags_plural.go
AnuiDolphin: Addressed in 5890487. Added `TestStackTagsPluralRead / Returns an error when stored inputs can't be decoded` — constructs inputs where `tags` is a string rather than a map, asserts `Read` returns the decode error with prefix `failed to decode stored inputs` rather than falling into adopt-all.

provider/pkg/pulumiapi/stack_tags.go
AnuiDolphin: Addressed in 5890487. `GetStackTag` now delegates to `GetStackTags`:

```go
func (c *Client) GetStackTag(ctx context.Context, stackName StackIdentifier, tagName string) (*StackTag, error) {
    tags, err := c.GetStackTags(ctx, stackName)
    if err != nil {
        return nil, fmt.Errorf("failed to get stack tag: %w", err)
    }
    ...
}
```

Signature unchanged, error prefix preserved for any downstream log consumers.

provider/pkg/resources/stack_tags_plural.go
AnuiDolphin: Addressed in 5890487. Dropped the `ToPulumiServiceStackTagsInput` helper entirely. The one test that used it now calls `util.FromPropertyMap` directly, which is what the resource's Create/Read/Update paths already use.

provider/pkg/resources/stack_tags_plural.go
AnuiDolphin: Addressed in 5890487. The import branch is now `managedTags = maps.Clone(allTags)` so the returned state doesn't alias the API-client-owned map.

provider/pkg/util/property_map.go
AnuiDolphin: Addressed in 5890487. `set` now returns `unsupported map type %s: only map[string]string is supported` for any non-`map[string]string` field instead of silently no-oping. `get` can't return an error without a signature change that would cascade through all callers, so I added a doc comment on `ToPropertyMap` calling out the restriction and pointing callers at `FromPropertyMap` for the error-returning behavior.

provider/pkg/provider/manual-schema.json
AnuiDolphin: Addressed in 5890487. Trimmed the description to a short prose summary (what it manages, the immutability-driven delete-and-recreate, the import-adopts-all behavior) and linked out to the registry docs for the full example.

provider/pkg/resources/stack_tags_plural.go
AnuiDolphin: Addressed in 5890487. Added a two-line comment above the `partial := input; partial.Tags = created` pair explaining the value-struct / swap-Tags pattern.

AnuiDolphin:
Addressed all 9 review comments in 5890487 (force-pushed). Each thread has an individual reply with the specific fix; summary below. Threads left unresolved for the reviewer.

| # | Severity | File | Fix |
|---|---|---|---|
| 1 | Important | `stack_tags_plural_test.go` | Added `TestStackTagsPluralUpdatePartialFailure` with mid-delete and mid-create subtests that assert `ErrorResourceInitFailed.Properties.Tags` reflects the actual live set at the failure point. |
| 2 | Important | `stack_tags_plural_test.go` | Added `TestStackTagsPluralRead / Import adopts every tag on the stack when there is no prior state` — exercises `Read` with `req.Inputs == nil`. |
| 3 | Important | `stack_tags_plural_test.go` | Added `TestStackTagsPluralRead / Returns an error when stored inputs can't be decoded` — malformed `tags` field now produces `failed to decode stored inputs` instead of falling into adopt-all. |
| 4 | Suggestion | `pulumiapi/stack_tags.go` | `GetStackTag` delegates to `GetStackTags`; signature unchanged, error prefix preserved. |
| 5 | Suggestion | `stack_tags_plural.go` | Dropped `ToPulumiServiceStackTagsInput`; the one test caller now uses `util.FromPropertyMap` directly. |
| 6 | Suggestion | `stack_tags_plural.go` | Import branch now `managedTags = maps.Clone(allTags)` — no more aliasing the API-client-owned map. |
| 7 | Suggestion | `util/property_map.go` | `set` returns `unsupported map type ...` for non-`map[string]string` fields; `ToPropertyMap` gained a doc comment covering the restriction on the `get` side. |
| 8 | Suggestion | `manual-schema.json` | Description trimmed to a short prose summary; links out to registry docs for the full import example. |
| 9 | Nitpick | `stack_tags_plural.go` | Two-line comment added above the `partial := input; partial.Tags = created` pair explaining the value-struct / swap-Tags pattern. |

davidemassarenti-optio3 (approved):

Issue #751, opened 2026-04-21
Spurious `replace` on `TeamEnvironmentPermission` when upgrading provider


aureq:
### Describe what happened

When upgrading the `@pulumi/pulumiservice` from `0.27.3` to `0.36.0` (`0.29.3` is when the problem is first noticeable), the resource `pulumiservice:index:TeamEnvironmentPermission` is marked for replacement despite no changes in the code. Furthermore, `maxOpenDuration` now shows as a diff with an empty value while 1) it is marked as optional and 2) the value isn't set in the code either.

### Sample program

```ts
import * as pulumi from "@pulumi/pulumi";
import * as pulumiservice from "@pulumi/pulumiservice";


export = async () => {
    const pulumiProvider = new pulumiservice.Provider("pulumi-provider", {
        accessToken: pulumi.secret("pul-xxx"), // 🔥 use a valid token
    });

    const pulumiTeam = new pulumiservice.Team("smart-baguette-team", {
        name: "smart-baguette-team",
        displayName: "zendesk-smart-baguette-team", // somehow the state contains ""
        organizationName: "menfin-demo",
        teamType: "pulumi",
    }, { provider: pulumiProvider, parent: pulumiProvider, ignoreChanges: ['members'] });

    const foo = new pulumiservice.TeamEnvironmentPermission("test-team-env-permission", {
        organization: "menfin-demo",
        team: pulumiTeam.name.apply(name => name!),
        project: "sandbox",
        environment: "require-approuval",
        permission: pulumiservice.EnvironmentPermission.Open,
    });
}
```

The commands to run are:
```bash
# phase 1 - downgrade and deploy using 0.27.3
npm rm --save @pulumi/pulumiservice
rm -rf node_modules/@pulumi/pulumiservice && npm install --save @pulumi/pulumiservice@0.27.3
p up --yes # ✅ resources are created
p pre # ✅ no changes
p pre --diff # ✅ no changes
```

```bash
# phase 2 - upgrade to latest and deploy using 0.36.0
npm rm --save @pulumi/pulumiservice
rm -rf node_modules/@pulumi/pulumiservice && npm install --save @pulumi/pulumiservice@0.36.0
p pre # ❌ pulumiservice:index:TeamEnvironmentPermission is marked for replacement
p pre --diff # ❌ pulumiservice:index:TeamEnvironmentPermission is marked for replacement
p up --yes # ❌ replacement is performed
```

### Log output

```
Previewing update (0050)

[...]

PR #760, opened 2026-04-22
Fix TeamEnvironmentPermission replacement on upgrade from 0.29.3-0.36.0

borisschlosser:
## Summary

Follow-up to #752 that closes a residual upgrade path the merged fix did not cover.

#752 fixed `Check()` and `Read()` so they no longer emit `maxOpenDuration: \"\"` when the user did not set the field. However, users who ran provider versions **0.29.3 through 0.36.0** already have state on disk with `maxOpenDuration: \"\"` baked into the resource's `OldInputs` — written there by the buggy `Check` path before the fix. On the first preview after upgrading to the post-#752 provider:

- the user's program (unchanged) still does not set `maxOpenDuration`,
- the new `Check()` correctly produces inputs **without** the key,
- `Diff()` compares `OldInputs` (has `maxOpenDuration: \"\"`) against `News` (no key) and reports the key as deleted,
- `changedKeys` is passed verbatim into `Replaces`, so the resource is force-replaced once more.

That's the exact failure mode #751 set out to eliminate, just on a slightly different upgrade path (0.29.3–0.36.0 → 0.37.0) than the one covered by the regression test in #752 (pre-0.29.3 → 0.29.3+).

## Fix

Normalize an empty-string `maxOpenDuration` to absent on **both sides** of the Diff comparison. Since the resource's semantics have always treated `\"\"` and unset as the same thing — `Create()` was already sending `nil` to the API for `\"\"` — making them compare equal in Diff is consistent and non-breaking. State converges to the clean representation on the next successful create/refresh without any user action.

```go
// team_environment_perm.go, Diff()
normalizeEmptyMaxOpenDuration(olds)
normalizeEmptyMaxOpenDuration(news)
```

### Why Diff and not Check

An alternative would be re-emitting `\"\"` in `Check()` when old state had it, to keep state bit-identical. That was rejected because:

1. It would require reading `req.GetOlds()` in Check, which this resource does not currently do.
2. It would preserve the bad `\"\"` in state forever, defeating the point of the #752 cleanup.
3. Making Check's output depend on state history is harder to reason about than a localized Diff-time normalization.

## Tests

Two new unit tests in `provider/pkg/resources/team_environment_perm_test.go`:

- **`TestTeamEnvironmentPermission_Diff_UpgradeFromEmptyStringMaxOpenDuration`** — mirrors the existing `TestTeamEnvironmentPermission_Diff_UpgradeFromPreMaxOpenDuration` regression test but with `OldInputs` containing `maxOpenDuration: \"\"` (the 0.29.3–0.36.0 shape). Asserts `DIFF_NONE` and no `Replaces`.
- **`TestTeamEnvironmentPermission_Diff_DetectsRealMaxOpenDurationChange`** — guards against over-aggressive normalization. A real value change (`30m0s` → `1h0m0s`) must still surface as a change **and** appear in `Replaces`, since this resource replaces on any change.

All existing tests from #752 continue to pass.

## Test plan

- [x] `go test -short -run TeamEnvironmentPermission ./provider/pkg/resources/...` — all 10 cases pass locally
- [x] CI integration tests (`TestYamlTeamEnvironmentPermissionExample`, etc.)
- [ ] Manual validation against a stack created on 0.29.3–0.36.0 without `maxOpenDuration` set:
  - `pulumi preview` on the upgraded provider should show no changes (previously: `+-TeamEnvironmentPermission` replace with `+ maxOpenDuration: \"\"`)

## Related

- Issue: #751
- Previous fix: #752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fixed TeamEnvironmentPermission refresh returning wrong permission when the same environment name exists in multiple projects [#674](https://github.com/pulumi/pulumi-pulumiservice/issues/674)
 - Fixed TeamEnvironmentPermission spurious replacement on upgrade from 0.29.2 caused by the optional `maxOpenDuration` field being serialized as an empty string in Check and Read [#751](https://github.com/pulumi/pulumi-pulumiservice/issues/751)
 - Fixed TeamEnvironmentPermission panic when `maxOpenDuration` was supplied as a non-string value; `Check` now returns a `CheckFailure` at preview instead of crashing during apply [#751](https://github.com/pulumi/pulumi-pulumiservice/issues/751)
+- Fixed TeamEnvironmentPermission spurious replacement for users upgrading from provider versions 0.29.3–0.36.0 whose state contains an empty-string `maxOpenDuration`; `Diff` now treats an empty-string `maxOpenDuration` as equivalent to an absent field [#751](https://github.com/pulumi/pulumi-pulumiservice/issues/751)
 
 ## 0.35.0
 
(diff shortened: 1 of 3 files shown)


seanyeh (approved):

aureq:
@borisschlosser this is a really good catch! 🎉

PR #762, opened 2026-04-23
Add RBAC resources: OrganizationMember, OrganizationRole, TeamRoleAssignment, role data sources

lukebward:
### New resources
- **`OrganizationRole`** — define custom fine-grained org roles with a `PermissionDescriptor` tree and `resourceType` (default `global`).
- **`TeamRoleAssignment`** — assign a custom role to a team; auto-enables the team's custom-roles feature on first use.
- **`OrganizationMember`** — manage a user's org membership and role (built-in `role` or custom `roleId`).

### New data sources
- **`getOrganizationMembers`** — list all org members with roles.
- **`getOrganizationMember`** — single-member lookup by `username` or `email` (case-insensitive); errors when not found. Closes [pulumi-service#41668](https://github.com/pulumi/pulumi-service/issues/41668).
- **`getOrganizationRoleScopes`** — discover the permission-scope catalogue so customers don't have to guess valid scope names.
- **`getCurrentUser`** — the user the provider's access token belongs to. Useful for seeding new `Team.members` with the creator (Pulumi Cloud auto-adds them, omitting causes refresh drift).
- **`getEnvironment`** — look up an existing ESC env by `(organization, project, name)` and get its UUID.

### New helper functions (builders, not data fetches)
- **`buildEnvironmentScopedPermissions` / `buildStackScopedPermissions` / `buildInsightsAccountScopedPermissions`** — build the `PermissionDescriptorGroup → Condition → LiteralExpression` tree for an entity-pinned role, no hand-rolled JSON. Each helper exposes a `permissions` Mapping and a `permissionsJson` string sibling (workaround for Pulumi's Python SDK stripping `__`-prefixed keys from invoke responses). Named `build*` rather than `get*` because they synthesize a descriptor from inputs — no service round-trip.
  - **TypeScript / YAML / Go / .NET / Java**: use the `permissions` field directly:
    ```ts
    permissions: service.buildEnvironmentScopedPermissionsOutput({
        environmentId: env.environmentId.apply(id => id!),
        permissions: ["environment:read", "environment:open"],
    }).permissions
    ```
  - **Python**: the structured Mapping is stripped of `__`-prefixed keys on the invoke path, so consume `permissionsJson` and `.apply(json.loads)` it instead:
    ```py
    permissions=build_environment_scoped_permissions_output(
        environment_id=env.environment_id,
        permissions=["environment:read", "environment:open"],
    ).permissions_json.apply(json.loads)
    ```

### New outputs / behaviors
- **`Environment.environmentId`** — the env's UUID, computed from `GET /api/esc/environments/{org}/{project}/{name}/metadata`. Required for `PermissionLiteralExpressionEnvironment.identity`.
- **`OrganizationMember`** adoption — if the user is already in the org, Create flips `adopted=true` and proceeds to set the role; Delete on an adopted member only un-assigns the custom role (snap back to built-in `member`), Delete on a non-adopted member removes the user.
- **`OrganizationRole.Read`** preserves unset `resourceType` so refresh doesn't drift on fields the program didn't set.
- **`OrganizationRole.Read`** surfaces JSON parse errors on `Details` instead of silently dropping the permission tree from state.
- **`OrganizationRole.Check`** tolerates computed/unknown inputs so `permissions` from a helper output (or an `apply()`) doesn't trip emptiness validation during preview.
- **`OrganizationRole.Delete` with `force=true`** revokes outstanding assignments before delete so dependent teardown doesn't block on a dangling role reference.
- **`AssignRoleToTeam`** uses POST (was wrong before).
- **Non-JSON API error body handling** — surfaces a readable error when the service returns HTML/plain-text.
- **Team custom-roles feature auto-enable** on first `TeamRoleAssignment`.

## Notes

- The `permissions` field on `OrganizationRole` is a JSON blob matching the service's permission-descriptor shape. The new helpers cover the common entity-pinned cases; a typed schema for the variants can follow once upstream codegen lands.
- `Environment` is still a legacy-style resource (raw protobuf), so its schema lives in `provider/pkg/provider/manual-schema.json`. New resources/data sources in this PR are infer-based — schema is generated from Go struct tags.
- Manually verified against a live org: Create for all three resources, the 409 adoption branch, all helpers, env-scoped role end-to-end.

## Test plan

- [x] `go test -short ./...` green
- [x] `go vet` clean (incl. examples with `-tags=all`)
- [x] Schema + all five SDKs regenerated
- [x] CI green on staging `service-provider-test-org`:
  - [x] `TestYamlRbacExample` — exercises the inline-JSON form + env-scoped role
  - [x] `TestRbacExample` (ts-rbac) — uses `buildEnvironmentScopedPermissionsOutput`; asserts no preview/refresh drift
  - [x] `TestPythonRbacExample` — uses `build_environment_scoped_permissions_output` via `permissions_json.apply(json.loads)` (the `__`-key-strip workaround)


examples/ts-rbac/index.ts
```diff
@@ -0,0 +1,79 @@
+import * as pulumi from "@pulumi/pulumi";
```
komalali, line 1: We should add this to examples_nodejs_test.go if we want this to run as a test

lukebward: I had this one as a manual test.  There is another example test there for the CI.

lblackstone:
### Code review

Found 5 issues:

1. `OrganizationRole.Update` always sends a non-nil `description` pointer (to `""`) even when the user did not set one. `UpdateRole`'s contract is "any of the three may be nil to leave unchanged," but `util.OrZero(core.Description)` returns `""` and `&desc` is then passed unconditionally. The `omitempty` tag on `*string` only omits nil, not pointers to empty strings, so the PATCH body includes `"description": ""` and clears any existing description on the server when updating only `name`.

https://github.com/pulumi/pulumi-pulumiservice/blob/35743cd7b4174c651b093e5b84a19ed4afea81da/provider/pkg/resources/organization_role.go#L202-L212

2. `OrganizationMember.Update` DryRun branch drops the `Adopted` flag. `Adopted` is a top-level field on `OrganizationMemberState` (not on `OrganizationMemberCore`), so the bare struct literal `OrganizationMemberState{OrganizationMemberCore: core}` returns `Adopted: false` for an adopted member. The non-DryRun path correctly carries it via `state.Adopted = req.State.Adopted`, and `Create` similarly preserves it on its DryRun path. `pulumi preview` on an adopted member will show a spurious `adopted: true → false` diff.

https://github.com/pulumi/pulumi-pulumiservice/blob/35743cd7b4174c651b093e5b84a19ed4afea81da/provider/pkg/resources/organization_member.go#L205-L211

3. The `getOrganizationMembers` data source surfaces `Role` inconsistently with the `OrganizationMember` resource. `memberInfoFrom` always sets `Role: m.Role` (the wire built-in role string), even when `m.FGARole` indicates a custom role is active. `applyMemberRoleToState` in the resource explicitly sets `state.Role = nil` for custom-role members. For the same member with a custom role, the data source returns `{role: "member", roleId: "<custom>"}` while the resource state has `{role: nil, roleId: "<custom>"}` — divergent shapes for the same underlying entity.

https://github.com/pulumi/pulumi-pulumiservice/blob/35743cd7b4174c651b093e5b84a19ed4afea81da/provider/pkg/functions/org_members.go#L188-L204

4. `ListOrgMembers` silently drops the `?type=backend` query parameter that the prior implementation sent on `master`. The new pagination loop omits it entirely and the test asserts `r.URL.Query().Get(\"type\") == \"\"`, so the removal is intentional — but it is undocumented in the PR description / CHANGELOG. If `type=backend` previously filtered out service-account members, this changes the shape of `ListOrgMembers` output (and therefore the new `getOrganizationMembers` data source and `GetOrgMember` lookups) for every consumer. Worth confirming with the service team and noting the behavior change.

https://github.com/pulumi/pulumi-pulumiservice/blob/35743cd7b4174c651b093e5b84a19ed4afea81da/provider/pkg/pulumiapi/members.go#L148-L178

5. `OrganizationMember.Delete` comment misdescribes what is sent over the wire. The comment says it PATCHes \"with only \`role\` set\" so the service \"converts legacy role input onto the backing FGA role.\" But `UpdateOrgMemberRole`'s own contract comment states the endpoint rejects built-in role names, and the implementation explicitly resolves built-ins to FGA IDs via `ResolveBuiltInRoleID` and sends `{fgaRoleId: ...}` instead. The behavior is correct; only the comment is wrong and could mislead future maintainers.

https://github.com/pulumi/pulumi-pulumiservice/blob/35743cd7b4174c651b093e5b84a19ed4afea81da/provider/pkg/resources/organization_member.go#L252-L267

provider/pkg/pulumiapi/roles.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	return out, nil
+}
+
+// ListOrgRoles returns the role catalogue for an organization filtered by
+// uxPurpose (e.g. "role", "policy"). The service requires uxPurpose — calling
+// the endpoint without it returns 400 Bad Request.
+func (c *Client) ListOrgRoles(ctx context.Context, orgName, uxPurpose string) ([]RoleDescriptor, error) {
+	if len(orgName) == 0 {
+		return nil, errors.New("organization name should not be empty")
```
lblackstone: nit: s/should/must since it errors

```suggestion
		return nil, errors.New("organization name must not be empty")
```

lblackstone:
A couple of nits from validating the API client against the production OpenAPI spec — both optional, nothing wire-breaking:

1. **Stale comment in `provider/pkg/pulumiapi/team_roles.go:83`** — the comment `// Service registers this as POST (despite the Java spec showing PUT).` no longer reflects the spec. The current Java IDL / generated `openapi.json` registers this endpoint as `POST` (`UpdateTeamRoles`), so the caveat is misleading now. Safe to drop.

2. **Consider explicit `json:` tags on `Member` and `User` in `provider/pkg/pulumiapi/members.go`** — the response fields (`Role`, `User`, `KnownToPulumi`, `VirtualAdmin`, `Name`, `GithubLogin`, `AvatarURL`, `Email`) currently rely on Go's case-insensitive JSON unmarshaling to line up with the spec's `role` / `user` / `knownToPulumi` / `virtualAdmin` / `name` / `githubLogin` / `avatarUrl` / `email`. It works today, but explicit tags would (a) survive a hypothetical move to a stricter decoder and (b) make the wire contract self-documenting next to the type. `FGARole` and `ContinuationToken` already have tags; the rest would match the same pattern.

Neither blocks. Good call on the `UpdateRole` lowercase-casing — the spec schema still shows PascalCase (`Name`/`Description`/`Details`) but the handler's custom `UnmarshalJSON` in `pkg/apitype/roles.go` only accepts lowercase, so your wire format is correct and matches reality.

---
_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

examples/examples_python_test.go
```diff
@@ -57,3 +57,36 @@ func TestPythonApprovalRulesExample(t *testing.T) {
...
+			"digits":           generateRandomFiveDigits(),
+		},
+		Dependencies: []string{
+			filepath.Join("..", "sdk", "python", "bin"),
+		},
+		// Enabling custom roles on the team causes the service to add the
+		// caller as a team member, which shows up on refresh as drift on
+		// the pre-existing Team resource. Same follow-up the yaml-rbac
+		// test is waiting on — not in scope here.
+		ExpectRefreshChanges: true,
```
komalali: @lukebward - I'm going to remove this - it's an indication that this example is going to have refresh drift. The tests are going to start failing. Will push a commit to fix.

provider/pkg/pulumiapi/team_roles.go
```diff
@@ -0,0 +1,140 @@
...
+	if _, err := c.do(ctx, http.MethodPost, apiPath, nil, nil); err != nil {
+		switch GetErrorStatusCode(err) {
+		case http.StatusConflict, http.StatusBadRequest:
+			// Already enabled — best-effort idempotency.
+			return nil
+		}
+		return fmt.Errorf("failed to enable team roles: %w", err)
+	}
+	return nil
+}
```
komalali: This is very much a side-effect. Should it be here?

provider/pkg/functions/current_user.go
```diff
@@ -0,0 +1,81 @@
...
+		"Returns the Pulumi Cloud user that the provider's access token belongs to. "+
+			"Useful for seeding a newly-created `Team` with the creator as a member, since "+
+			"Pulumi Cloud auto-adds the creator and omitting them causes a refresh drift.",
+	)
+	a.SetToken("index", "getCurrentUser")
+}
+
+func (o *GetCurrentUserOutput) Annotate(a infer.Annotator) {
+	a.Describe(&o.Username, "The user's Pulumi Cloud username (same as `githubLogin`).")
+	a.Describe(&o.GithubLogin, "The user's GitHub login.")
```
fallimic: Do we really want to expose GithubLogin? Here and in the comment? This is really an internal naming convention for us even if they don't use github.

provider/pkg/functions/org_members.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// OrganizationMemberInfo carries the fields returned by the backend member
+// roster for Pulumi Cloud organizations. Display name and email are not
+// included: the backing identity provider (GitHub/GitLab/Bitbucket) does not
+// populate them in the member list, and Pulumi only knows those fields for
+// members who have already signed in (surfaced via `knownToPulumi`).
+type OrganizationMemberInfo struct {
+	Username      string  `pulumi:"username"`
+	GithubLogin   string  `pulumi:"githubLogin"`
```
fallimic: Same here

provider/pkg/resources/organization_role.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	if core.ResourceType == nil && role.ResourceType != "" && role.ResourceType != "global" {
+		core.ResourceType = util.OrNil(role.ResourceType)
+	}
+	if core.UxPurpose == nil && role.UXPurpose != "" && role.UXPurpose != "role" {
+		core.UxPurpose = util.OrNil(role.UXPurpose)
+	}
+	if len(role.Details) > 0 {
+		parsed := map[string]interface{}{}
+		if err := json.Unmarshal(role.Details, &parsed); err == nil {
```
danbiwer: do we need to return this error?

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1952,6 +2013,165 @@
...
+          "description": "The role's permission descriptor tree — passed to the service verbatim. This is the `details` field of a Pulumi Cloud PermissionDescriptor: an object with a `__type` discriminator (e.g. `PermissionDescriptorAllow`, `PermissionDescriptorCompose`) describing which scopes are granted."
+        },
+        "resourceType": {
+          "type": "string",
+          "description": "The resource type the role's permissions apply to. Defaults to `global` (the org-wide role that can be assigned to members and teams). Other valid values: `stack`, `environment`, `insights-account`."
+        },
+        "uxPurpose": {
+          "type": "string",
+          "description": "How the role appears in the Pulumi Cloud console. One of `role`, `role_private`, `policy`, `set`. Defaults to `role`."
+        }
```
komalali: This seems like a very strage property name. What's the actual intent here?

lukebward: Hmmm, this leaked from the actual API.  I think we can kill it here

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -1952,6 +2013,165 @@
...
+          "type": "string",
+          "description": "The Pulumi Cloud organization name.",
+          "replaceOnChanges": true
+        },
+        "permissions": {
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "pulumi.json#/Any"
+          },
+          "description": "The role's permission descriptor tree — passed to the service verbatim. This is the `details` field of a Pulumi Cloud PermissionDescriptor: an object with a `__type` discriminator (e.g. `PermissionDescriptorAllow`, `PermissionDescriptorCompose`) describing which scopes are granted."
```
komalali: Not sure this will be satisfactory to explain what's being allowed here? Should we link to docs?

provider/cmd/pulumi-resource-pulumiservice/schema.json
lukebward: It's used to filter roles by "org, team, token" etc

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -688,6 +688,40 @@
       },
       "type": "object"
     },
+    "pulumiservice:index:OrganizationMemberInfo": {
+      "properties": {
+        "knownToPulumi": {
+          "type": "boolean",
+          "description": "Whether this member has a Pulumi Cloud account."
+        },
```
komalali: Does this have value to the end user?

komalali: Maybe yes?

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -2437,6 +2657,58 @@
...
+          "replaceOnChanges": true
+        },
+        "roleId": {
+          "type": "string",
+          "description": "The ID of the custom role to assign.",
+          "replaceOnChanges": true
+        },
+        "roleName": {
+          "type": "string",
+          "description": "The name of the assigned role at the time of last refresh."
```
komalali: "at the time of last refresh" seems odd?

provider/cmd/pulumi-resource-pulumiservice/schema.json
lukebward: Maybe, but is it something that we need to expose?

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -2756,6 +3028,39 @@
     }
   },
   "functions": {
+    "pulumiservice:index:getCurrentUser": {
+      "description": "Returns the Pulumi Cloud user that the provider's access token belongs to. Useful for seeding a newly-created `Team` with the creator as a member, since Pulumi Cloud auto-adds the creator and omitting them causes a refresh drift.",
```
komalali:
```suggestion
      "description": "Returns the Pulumi Cloud user that the provider's access token belongs to. Useful for seeding a newly-created `Team` with the creator as a member, since Pulumi Cloud auto-adds the creator. Omitting this user from the team definition will result in a refresh drift.",
```

provider/cmd/pulumi-resource-pulumiservice/schema.json
lukebward: This is the bigger issue here.  Because of the union codegen, we currently would need to use this nasty json which isn't really documented
```ts
const scopedReadOnlyRole = new service.OrganizationRole("scopedReadOnlyRole", {
    organizationName,
    name: `ts-rbac-scoped-read-only-${nameSuffix}`,
    description: "Read+open access scoped to a single ESC environment.",
    permissions: scopedEnv.environmentId.apply(envId => ({
        __type: "PermissionDescriptorGroup",
        entries: [
            {
                __type: "PermissionDescriptorCondition",
                condition: {
                    __type: "PermissionExpressionEqual",
                    left: { __type: "PermissionExpressionEnvironment" },
                    right: {
                        __type: "PermissionLiteralExpressionEnvironment",
                        identity: envId,
                    },
                },
                subNode: {
                    __type: "PermissionDescriptorAllow",
                    permissions: ["environment:read", "environment:open"],
                },
            },
        ],
    })),
});```

Comment (davidemassarenti-optio3) [reply provider/cmd/pulumi-resource-pulumiservice/schema.json]: For the time being, could we hide it in a set of helper functions? One for envs, one for accounts, one for stacks. The inputs would be entity id and the set of permissions to grant.

File: provider/pkg/provider/manual-schema.json [hunk 13]
```diff
@@ -1777,6 +1777,10 @@
         "revision": {
           "description": "Revision number of the latest version.",
           "type": "integer"
+        },
+        "environmentId": {
+          "description": "The environment's UUID. Use this as the `identity` value when pinning a custom RBAC role to this environment via a `PermissionLiteralExpressionEnvironment` expression in `OrganizationRole.permissions`.",
+          "type": "string"
```
komalali, line 1783: What is this file? How is this different than provider/schema.json?

lukebward: I *guess* this is legacy schema, don't know if there's a good reason for maintaining.  We don't actually use this as a build artifact

provider/cmd/pulumi-resource-pulumiservice/schema.json
```diff
@@ -2756,6 +3014,133 @@
...
+        "required": [
+          "organizationName",
+          "projectName",
+          "name",
+          "environmentId"
+        ],
+        "type": "object"
+      }
+    },
+    "pulumiservice:index:getEnvironmentScopedPermissions": {
```
davidemassarenti-optio3: Maybe rename to buildEnvironmentScopedPermissions? Same for the other helpers

provider/pkg/provider/manual-schema.json
komalali: So this code change shouldn't be here / has no effect right?

komalali (approved):

provider/pkg/provider/manual-schema.json
lukebward: Not in this repo, but unsure if we use it anywhere else.

komalali: Nah this schema was only ever used here. I think it's just from before we migrated to the go sdk

PR #763, opened 2026-04-23
Replace hand-coded v1 with a generated OpenAPI-driven v2 provider

joeduffy:
v1 was hand-coded, and that model couldn't keep up with the growth of the Pulumi Cloud API. Entire domains had no path into the provider because every new resource meant hundreds of lines of bespoke Go. v2 inverts the relationship: Pulumi Cloud's OpenAPI specification is the source of truth, and the provider is generated from it. The only editable surface is a single mapping file that classifies each Pulumi Cloud operation as a Pulumi resource, function, method, or a deliberate exclusion with a written reason. A coverage gate fails CI if anything is left unclassified, so the mapping can't silently drift as Pulumi Cloud evolves. Keeping the provider current is now a weekly mechanical refresh rather than a standing tax on the team.

The core commitment is zero hand-coded resources. Where v1 reached for Go whenever a resource didn't fit one-op-per-verb CRUD, v2 instead extends the metadata schema with declarative primitives expressive enough to cover every such case. The rule is explicit in the maintainer docs: if a new Pulumi Cloud endpoint can't be expressed today, the right answer is to add another primitive, not to reintroduce an escape hatch. The discipline is load-bearing for scale — the alternative is a slow slide back into bespoke per-resource code.

The user-visible shape changes too. The flat pulumiservice:index:* namespace becomes sub-modules that mirror the URL structure, so the provider scales to hundreds of resources without a sea of peers. Type names drop redundant prefixes that only existed because the flat namespace offered no context. Pulumi state itself isn't touched by the upgrade — only the client-side tokens change — and the rename table plus pulumi state rename recipes ship alongside the code.

v2 covers v1 feature parity plus two preview-backed resources that v1 couldn't ship, across all five language SDKs. Twelve canonical end-to-end scenarios serve as both the integration test suite and the design review artifact for the sub-module partition; three of them are hand-written in every language as showcases of the distinct SDK idioms, the rest reachable via pulumi convert. A long-form maintainer playbook captures the refresh workflow end-to-end, including a Claude Code prompt that executes the weekly spec sync and triage.


joeduffy:
@claude Review this.

lukebward:
I'm going to try a new `ultrareview` here and see what we get.

EDIT: I lied, Anthropic is crashing right now on that.  Time to use the normal code-review.

lukebward:
### Code review

Found 4 issues:

1. `Call` / `Invoke` RPCs are not implemented, yet the CHANGELOG advertises `Account.triggerScan` method and `listAccounts` data source as shipped in 2.0.0-alpha.1. The new `Server` only embeds `pulumirpc.UnimplementedResourceProviderServer`, so both RPCs will return "not implemented" at runtime. Either move these features to the deferred section or wire up `Call`/`Invoke` through the dispatcher.

https://github.com/pulumi/pulumi-pulumiservice/blob/1898d13fb1c5fc39c9364cc77214ab174789f8b0/provider/pkg/provider/provider.go#L27-L34

2. OIDC Issuer `policies` uses `sortOnRead: true`, which re-introduces the spurious-diff bug fixed in [#542](https://github.com/pulumi/pulumi-pulumiservice/pull/542) (commit 5cbebfe). `sortPropertyArray` in `ResponseToOutputs` sorts the server response but `InputsToRequest` does not sort user input, so any order mismatch between what the user writes and what the service returns produces a perpetual diff.

https://github.com/pulumi/pulumi-pulumiservice/blob/1898d13fb1c5fc39c9364cc77214ab174789f8b0/provider/resource-map.yaml#L208-L216

3. Access-token `value` is dropped from state on first refresh. `OrgAccessToken` / `TeamAccessToken` / `AccessToken` read via `ListOrgTokens` / `ListTeamTokens`, which do not return `tokenValue`. For a property with `source: response, secret: true, output: true`, `ResponseToOutputs` has no fallback to preserve it from prior state — `diffMode: ciphertext`, `WriteOnly`, and `source: path|query` fallbacks all miss — so the secret value is silently omitted after the first `pulumi refresh`. The v1 resources explicitly restored this from prior state; the v2 metadata needs an equivalent (e.g. `diffMode: ciphertext` or a new preserve-from-state flag).

https://github.com/pulumi/pulumi-pulumiservice/blob/1898d13fb1c5fc39c9364cc77214ab174789f8b0/provider/resource-map.yaml#L132-L138

4. `docs/UPGRADE-v1-to-v2.md` claims "Resource IDs keep their format and value", but at least four resources change ID format: `OrgAccessToken` (3 parts → 2), `TeamAccessToken` (4 → 3), `StackTags` (4 → 3, no trailing `tags` segment), and `ApprovalRule` (`environment/org/project/env/ruleID` → `{organizationName}/{gateId}`). Users following the documented `pulumi state rename --type-token` recipe will hit `DecomposeID` regex mismatches on the first `up`/`refresh` for these resources — the guide needs to document `pulumi state rename --id` (or equivalent) for them.

https://github.com/pulumi/pulumi-pulumiservice/blob/1898d13fb1c5fc39c9364cc77214ab174789f8b0/docs/UPGRADE-v1-to-v2.md#L186-L191

iwahbe (commented):
If we are serious about this approach, I'd recommend a couple of major changes to the implementation:

1. We should stay on `github.com/pulumi/pulumi-go-provider` instead of dealing with raw gRPC. Only `github.com/pulumi/pulumi-go-provider/infer` requires hand-writing go-code. All the deeper layers don't require that, but they provide a much nicer API to work on then the gRPC and we can then still take advantage of other layers (cancel and routing come to mind here).

2. A generator binary + a runtime binary makes development much worse and adds complexity for no real benefit. It is IMO a premature optimization. Instead, we should directly embed the OpenAPI spec & metadata.json in the runtime binary, and have it build the schema on an as-needed bases (erroring on `GetSchema` if the mapping is incomplete or incorrect). This simplifies development, since `go build` continues to work as expected, instead of needing a multi-step build. This also gives us some cool runtime abilities, like the ability to use parameterization to allow trialing new OpenAPI schemas or metadata files on-demand. This would allow self-hosted customers to generate a provider that fits their version of the service without building any Go code.

iwahbe:
I'd also do this as a more gradual cut-over. Merge in the mechanism for the v2 system as a new layer on top of the existing `github.com/pulumi/pulumi-go-provider.Provider` and use that to add new resources. We can merge that immediately to see all missing resources added, and also give ourselves some time to test the new resources before we switch over everything for v2. That gives us v2 as a time to make breaking changes to our OpenAPI derived schema.

lukebward:
> If we are serious about this approach, I'd recommend a couple of major changes to the implementation:
> 
> 1. We should stay on `github.com/pulumi/pulumi-go-provider` instead of dealing with raw gRPC. Only `github.com/pulumi/pulumi-go-provider/infer` requires hand-writing go-code. All the deeper layers don't require that, but they provide a much nicer API to work on then the gRPC and we can then still take advantage of other layers (cancel and routing come to mind here).
> 2. A generator binary + a runtime binary makes development much worse and adds complexity for no real benefit. It is IMO a premature optimization. Instead, we should directly embed the OpenAPI spec & metadata.json in the runtime binary, and have it build the schema on an as-needed bases (erroring on `GetSchema` if the mapping is incomplete or incorrect). This simplifies development, since `go build` continues to work as expected, instead of needing a multi-step build. This also gives us some cool runtime abilities, like the ability to use parameterization to allow trialing new OpenAPI schemas or metadata files on-demand. This would allow self-hosted customers to generate a provider that fits their version of the service without building any Go code.

I am vibing a prototype of what you suggested Ian.  So far, it's more legible.  I will link a PR when it's in a reviewable state.

PR #778, opened 2026-04-28
Rename `OrganizationRole.permissions` discriminator: `__type` → `kind`

komalali:
## Summary

- Reshapes `OrganizationRole.permissions` to **two kinds plus an optional `on:` modifier**. Descriptors are `{kind: "allow", permissions: [...]}` (terminal grant) or `{kind: "group", entries: [...]}` (composition). Either may carry `on: {environment: <uuid>}` / `on: {stack: <id>}` / `on: {insightsAccount: <id>}` to scope the grant to a single entity.
- The provider translates `on:` into the Pulumi Cloud REST API's wire shape `Condition(Equal(Expression<E>, Literal<E>(id)), <inner>)` internally on Create/Update, and collapses any `Condition(Equal(...))` back into an `on:` modifier on Read. **Wire format unchanged.**
- Drops the Python-only `permissionsJson` sibling output on the three `buildXScopedPermissions` helpers and the `apply(json.loads)` workaround it required. Python users now consume `.permissions` directly like every other language.
- Helpers radically simplify: each returns a flat `{kind: "allow", on: {entity: id}, permissions: [...]}` directly. No more single-entry-Group wrapper.
- Pre-release polish on the unreleased RBAC API ([#762](https://github.com/pulumi/pulumi-pulumiservice/pull/762)). Not a breaking change for any released user.

## Why

Originally noticed in [examples/py-rbac/__main__.py:130-140](https://github.com/pulumi/pulumi-pulumiservice/blob/main/examples/py-rbac/__main__.py#L130-L140), where the example needed a long Python-only workaround to round-trip `__type` discriminators through `permissionsJson` because Pulumi's Python SDK strips `__`-prefixed keys from invoke responses ([upstream issue: pulumi/pulumi#22738](https://github.com/pulumi/pulumi/issues/22738)).

Two paths to fix it:
1. Wait for the upstream Python SDK fix.
2. Stop exposing `__type` (and the structural complexity around it) at our SDK boundary at all.

This PR is path 2 — cleaner, faster to ship, and removes an asymmetry users in every language could see.

The redesign happened in two phases:
- **Phase 1 (early commits):** Rename `__type` → `kind` with 10 short variant names (`allow`/`group`/`condition`/`equal`/three `expression*` + three `literal*`).
- **Phase 2 (final commits — this is the cleanup):** Collapse the structural complexity into a 2-kind + 1-modifier shape. The `condition`/`equal`/`expression*`/`literal*` machinery exists today only on the wire; users see `kind: "allow"` + `on: { environment: <uuid> }` and the provider expands.

Phase 2 (typed discriminated unions) was considered and deferred: research showed Go/.NET/Java codegen erodes `oneOf`+`discriminator` to `object`, so the typing wins would be TS-only and the maintenance overhead (recursive shapes force inlining at every use site, per [pulumi/pulumi#14674](https://github.com/pulumi/pulumi/issues/14674)) wasn't worth it for now.

## What's in the diff

- `provider/pkg/resources/permission_descriptor.go` — translator does structural transformation (not just key-rename). Forward path expands `on:` into a Condition wrapper; reverse path collapses any `Condition(Equal(...))` back into `on:`. Errors on nested Conditions to avoid silent data loss.
- `provider/pkg/resources/organization_role.go` — `Check` validates `on:` shape up front (rejects empty / multi-key / unknown entity / non-string identity) so preview surfaces a clear error instead of a 400 at apply.
- `provider/pkg/functions/role_permissions.go` — three helpers' `Invoke` is six lines each. Drops the prior `Group → Condition → Equal → Expression/Literal → Allow` construction and the six `kindXxx` constants.
- `examples/{yaml-rbac,ts-rbac,py-rbac}` — examples reflect the simpler shape. yaml-rbac's env-scoped role drops 10 lines.
- `provider/cmd/pulumi-resource-pulumiservice/schema.json` + all five SDKs regenerated — only docstrings change in the SDKs (the underlying `Mapping[str, Any]` type stays the same, so cross-language signatures are stable).
- `CHANGELOG.md` — `## Unreleased / ### Improvements` updated.

## Test Plan

Translator coverage (provider/pkg/resources/permission_descriptor_test.go):
- 5 cardinal-shape fixtures (`flatAllow`, `scopedAllow`, `flatGroup`, `scopedGroup`, `mixedGroup`) × forward + reverse + round-trip in both directions.
- Per-entity positive cases for `stack` and `insightsAccount`.
- 13 forward-error subtests (every `on:` validation failure mode + every malformed Allow/Group case).
- 7 reverse-error subtests (malformed wire shape, mismatched Equal pair, etc.).
- 1 nested-Condition rejection (silent-overwrite guard).
- 1 legacy single-entry-Group wire shape (proves an in-flight PR-tester's state Reads cleanly).

Helper unit tests (provider/pkg/functions/role_permissions_test.go):
- 9 subtests across three helpers, each verifying the new `{kind: "allow", on: {entity: id}, permissions: [...]}` output shape.

Resource tests (provider/pkg/resources/organization_role_test.go):
- 3 new Check subtests pin the `on:` validation contract (empty / multi-key / unknown entity).

- [x] `cd provider/pkg && go test -short -count=1 -timeout 5m ./...` — 173/173 pass
- [x] `grep -rn "kind: condition\|kind: equal\|kind: expressionEnvironment\|kind: literalEnvironment" provider/ sdk/ examples/ CHANGELOG.md` — empty
- [x] `grep -rn "permissionsJson\|permissions_json" sdk/ examples/ provider/` — empty
- [x] No commit in this final batch except `Regenerate SDKs:` touched `sdk/` (CI-friendly per the build_sdks rule)
- [ ] `cd examples && go test -tags yaml -run TestYamlRbacExample -timeout 10m`
- [ ] `cd examples && go test -tags nodejs -run TestTsRbacExample -timeout 15m`
- [ ] `cd examples && go test -tags python -run TestPyRbacExample -timeout 15m`

## Notes for reviewers

- The wire format on the Pulumi Cloud REST API is unchanged. Existing role state in the service round-trips cleanly: Read translates wire → kind, Create translates kind → wire, refresh produces no drift. `provider/pkg/pulumiapi/roles_test.go`'s wire-format fixture is intentionally still `__type`-shaped.


provider/pkg/resources/permission_descriptor.go
```diff
@@ -0,0 +1,329 @@
...
+	},
+}
+
+// wireToEntityType is the reverse map, built from entityTypeToWire so the two
+// stay in lockstep. Keyed by the (expression, literal) pair concatenated with
+// a separator unlikely to appear in either name.
+var wireToEntityType = func() map[string]string {
+	m := make(map[string]string, len(entityTypeToWire))
+	for entity, pair := range entityTypeToWire {
+		m[pair.expression+"|"+pair.literal] = entity
```
lukebward, line 51: is this kinda finnicky?

lukebward (approved):

PR #784, opened 2026-04-29
Add Stack.configEnvironment for Service-Backed Configuration

rgharris:
This has been tested in production.

Lets Stack opt into linking an ESC environment as its remote config and secrets store. One mode for now:

- managed: true has the server create a dedicated env named `<projectName>/<stackName>` inline at stack creation time, deleted with the stack via preserveEnvironment=false.

Potential future mode can be added:
- environment+project (with optional version) references an existing env; PSP doesn't own its lifecycle and passes preserveEnvironment=true on delete.

Closes pulumi/pulumi-service#41218


nyobe (approved):

provider/pkg/resources/stack.go
```diff
@@ -46,13 +153,59 @@ func (s *PulumiServiceStackResource) ToPulumiServiceStackTagInput(
...
+		// own — is a non-destructive PUT/DELETE on /config.
+		return plugin.DiffUpdate, true
+	}
+
+	// Both sides set with matching auto flag: env identity is replace-only
+	// because UpdateStackConfigHandler rejects re-linking to a different env.
+	if oldEnv.resolveEnvProject() != newEnv.resolveEnvProject() ||
+		oldEnv.Environment != newEnv.Environment {
+		return plugin.DiffUpdateReplace, true
+	}
```
nyobe: Won't this trigger replacement of the whole stack?

I thought the re-linking prohibition required deleting and recreating the _stack config_, not the whole stack

rgharris: fixed in 9aaa210

rgharris:
Closing in favor of https://github.com/pulumi/pulumi-service/issues/43936

PR #788, opened 2026-04-30
Import Role Fix

lukebward:
## Summary

Extends the `__type` → `kind` permissions-discriminator rename (#778) so externally-authored OrganizationRoles round-trip through `pulumi import`, plus adjacent fixes caught during the work.

**Core**
- `OrganizationRole.permissions` now passes wire-grammar shapes (`PermissionDescriptorCompose`, `Permission`, `PermissionExpression*`) through unchanged — UI- or REST-authored Compose roles import cleanly with `kind:` instead of the previous "unknown `__type`" error.
- Drop `buildTeamScopedPermissions`: team↔role association lives in `TeamRoleAssignment`, not a permission descriptor.
- `OrganizationRole.Read` validates `uxPurpose`; `OrganizationRole.Delete` tries an unprivileged delete first, escalates to `force` only on 409.
- Schema + SDKs (Go, .NET, Java, Node, Python) regenerated.

**Adjacent fixes**
- Provider config falls back to `PULUMI_BACKEND_URL` when the engine sends an empty `apiUrl` (self-hosted backends).
- `ListOrgMembers` merges backend + frontend rosters and filters the synthetic org-as-user entry.

**Test infrastructure**
- New `TestYamlRbacComposeImport` provisions a Compose role via direct REST, runs `pulumi import` against an empty stack, and asserts the generated YAML carries `kind: PermissionDescriptorCompose`.
- Three Insights examples switched from `fn::open::aws-login` (placeholder ARN) to static OIDC env vars to satisfy the new server-side ESC-credential validation in [pulumi-service#41970](https://github.com/pulumi/pulumi-service/pull/41970).

## Test plan

- [x] Provider unit tests pass
- [x] `TestYamlRbacComposeImport` green
- [x] Full acceptance test suite green on staging
- [x] Schema diff + SDK regen confirmed against mise-pinned Pulumi version


provider/pkg/resources/permission_descriptor_test.go
```diff
@@ -710,3 +712,145 @@ func mustJSON(t *testing.T, v interface{}) string {
...
+//     → wire JSON → kind), which round-trips through the generated
+//     UnmarshalJSONPermissionDescriptor. The generated permissionDescriptor
+//     AllowImpl only models `permissions` and `constraints` — anything else
+//     is silently dropped.
+func envScopedAllowsKind(envIDs []string) map[string]interface{} {
+	entries := make([]interface{}, 0, len(envIDs))
+	for _, id := range envIDs {
+		entries = append(entries, map[string]interface{}{
+			"kind":        "PermissionDescriptorAllow",
+			"on":          map[string]interface{}{"environment": id},
```
davidemassarenti-optio3: @lukebward , we need to remove "on" support. Use helper functions to construct higher-level roles, but the provider should just take a PermissionDescriptor.

komalali: This will break users, but there are only a few so far. We'll need to tell/show them the alternative.

provider/pkg/resources/permission_descriptor.go
```diff
@@ -12,178 +12,86 @@
...
-//   - If the user authored a top-level Group, do not collapse — preserve
-//     their Group(Condition) shape verbatim.
-//   - All other prior shapes (Allow, Compose, IfThenElse, …) cannot produce
-//     a single-entry Group(Condition) on the wire, so the gate has no
-//     effect on them.
-func permissionsFromWire(
+//   - If the user authored a top-level Condition (or imported a role
+//     with no prior input), collapse — the wrapped Group(Condition) wire
+//     shape reads back as Condition. Matches helper output.
+//   - If the user authored a top-level Group, do not collapse —
```
davidemassarenti-optio3: We should see if we can remove all these manipulations.

provider/pkg/functions/role_permissions.go
```diff
@@ -16,33 +16,172 @@ package functions
...
-const scopedPermissionsHelpDoc = "The result is directly assignable to " +
-	"`OrganizationRole.permissions`. To grant scopes on more than one entity " +
-	"in a single role, hand-roll a `group` whose `entries` list pulls the " +
-	"output of each helper."
+// rbacPermissionSlice converts a []string of scope names to the typed
+// apitype.RbacPermissionSlice the apitype builders consume.
+func rbacPermissionSlice(scopes []string) apitype.RbacPermissionSlice {
+	out := make(apitype.RbacPermissionSlice, len(scopes))
+	for i, s := range scopes {
+		out[i] = apitype.RbacPermission(s)
```
davidemassarenti-optio3: We should check that the value is a valid permission. Can we change this to return `(apitype.RbacPermissionSlice, error)`?

provider/pkg/pulumiapi/roles.go
```diff
@@ -15,24 +15,25 @@ package pulumiapi
...
 	"path"
+
+	"github.com/pulumi/pulumi-pulumiservice/provider/pkg/apitype"
 )
 
 type RoleClient interface {
-	CreateRole(ctx context.Context, orgName string, req CreateRoleRequest) (*RoleDescriptor, error)
-	GetRole(ctx context.Context, orgName, roleID string) (*RoleDescriptor, error)
+	CreateRole(
+		ctx context.Context, orgName string, req apitype.PermissionDescriptorBase,
```
davidemassarenti-optio3: If line is too long, best to split each parameter on a separate line

provider/pkg/resources/permission_descriptor.go
```diff
@@ -12,335 +12,27 @@
...
-	return map[string]interface{}{entityType: identity}, nil
+// validatePermissions sanity-checks a user-supplied descriptor map. The
+// descriptor variants themselves are opaque to the provider (Allow,
+// Group, Condition, Compose, IfThenElse, Select, the boolean operators,
+// and any future variant Pulumi Cloud adds pass through unchanged); we
+// only verify the top-level object carries the required `__type`
+// discriminator key, so users see a clear error at preview rather than a
+// 400 at apply.
+func validatePermissions(node map[string]interface{}) error {
+	if _, has := node["__type"]; !has {
```
davidemassarenti-optio3: We should just roundtrip this through JSON, don't assume any structure.

davidemassarenti-optio3 (approved):

PR #802, opened 2026-05-04
Migrate the next 4 resources: `TlsSchedule`, `DriftSchedule`, `EnvironmentRotationSchedule` and `DeploymentSchedule`

iwahbe:
Each commit is a single migration, and each commit leaves the provider in a valid state. The migrated resources are compatible at the schema & state level, meaning users won't need to migrate.

Stacked on top of https://github.com/pulumi/pulumi-pulumiservice/pull/800


provider/pkg/resources/schedules_test.go
lukebward, line 1: No more test for any schedule?

iwahbe: We still have integration tests. I can put back the test for Reading DeploymentSchedule, but I don't think it provides much protection. It's testing a lot less then it used to, since we no longer implement our own serde.

lukebward (approved):

fallimic (approved):

PR #807, opened 2026-05-05
[Preview] PSP v2 from OpenAPI spec

lukebward:
Adds a v2 resource namespace at `pulumiservice:v2:*` (accessed as `pulumiservice.v2.*` in user code) that derives resources from the embedded Pulumi Cloud OpenAPI spec via a metadata-driven layer with a dynamic CRUD handler. Existing v1 resources at `pulumiservice:index:*` continue to work unchanged. Includes per-language examples (yaml/typescript/python/go/csharp/java) and a unified test driver for the new namespace, plus CI wiring to run the v2 tests.


provider/pkg/provider/provider.go
```diff
@@ -167,6 +195,84 @@ func MakeProvider(host *provider.HostClient, name, version string) (pulumirpc.Re
...
+		return nil, fmt.Errorf("authedTransport: parse base URL %q: %w", t.baseURL, err)
+	}
+
+	req.URL.Scheme = base.Scheme
+	req.URL.Host = base.Host
+	req.Host = base.Host
+
+	req.Header.Set("Authorization", "token "+t.token)
+	if req.Header.Get("Accept") == "" {
+		req.Header.Set("Accept", "application/vnd.pulumi+8")
```
lblackstone: This version is now advertised by the service under the capabilities API. We'll likely want to parameterize this version and set it when the provider is released.

provider/pkg/rest/resource.go
```diff
@@ -0,0 +1,838 @@
...
+	if needsBody(op.Method) {
+		switch op.RequestContentType {
+		case "application/x-yaml":
+			// Raw-string body from the "yaml" input; absent leaves the body empty.
+			if v, ok := inputs.GetOk("yaml"); ok && v.IsString() {
+				body = strings.NewReader(v.AsString())
+				contentType = "application/x-yaml"
+			}
+		default:
+			bodyJSON, err := json.Marshal(propertyMapToAny(inputs))
```
fallimic: Don't we need to do a translation here back to the wire format? This will take any inputs and serialize them directly to JSON but the input names might not match the wire format. Additionally, path parameters/query parameters get added to the body in this case which probably won't cause issues but the server could reject the extra fields.

provider/pkg/rest/schema.go
```diff
@@ -0,0 +1,598 @@
...
+		if len(allowlist) > 0 {
+			if _, ok := allowlist[name]; !ok {
+				continue
+			}
+		} else if _, blocked := denylist[name]; blocked {
+			continue
+		}
+		ps := openAPIToProperty(p)
+		applyFieldMeta(&ps, rm.Fields[name], false)
+		if looksSecret(name) {
```
fallimic, line 331: Should we have a corresponding check on operationInputs?

provider/pkg/rest/resource.go
```diff
@@ -0,0 +1,838 @@
...
+	}
+	if op == nil {
+		return p.UpdateResponse{}, fmt.Errorf("update: resource has no update operation declared")
+	}
+	if req.DryRun {
+		return p.UpdateResponse{Properties: req.Inputs}, nil
+	}
+	src := mergeMaps(req.Inputs, req.OldInputs, req.State)
+	_, state, err := r.execAndDecode(ctx, op, src)
+	if err != nil {
```
fallimic, line 633: We probably need to do a read-after-update similar to read-after-create flow for the same reason, no?

lukebward: Yep, adding this in

provider/pkg/rest/schema.go
lukebward: Yeah definitely

lukebward:
### Code review

Found 3 issues:

1. `install_java_sdk` is now an empty target. The PR moved `publishToMavenLocal` into the `build_java` sentinel, but `test.yml` runs `make --touch ... build_sdks` (which fakes the sentinel) then `make install_sdks` — so the Java SDK is never published to local Maven and v1 Java tests like `TestJavaTeamsExamples` will fail to resolve `com.pulumi:pulumiservice`.

https://github.com/pulumi/pulumi-pulumiservice/blob/48f156135cddc92865da748c6abad61b69a6ef57/Makefile#L215-L220

2. Stale `idField` references in scaffolder docs and `encodeStable` preferred-key list — the actual struct field is `IDFormat string \`json:"idFormat"\`` (see `provider/pkg/rest/metadata.go:69`). The doc + `_note` in `provider/pkg/cloud/metadata.json:4` and `provider/pkg/cloud/spec.go:37` repeat the same wrong name. Worth a global rename to `idFormat`.

https://github.com/pulumi/pulumi-pulumiservice/blob/48f156135cddc92865da748c6abad61b69a6ef57/provider/tools/scaffold-metadata/main.go#L17-L20

3. The `Accept: application/vnd.pulumi+8` fallback in `authedTransport` is dead for v2 calls because `rest/resource.go` always sets `Accept: application/json` before the transport runs. This silently diverges from the project's documented header convention for v1 calls. Either drop the fallback in `provider.go:232-234` or stop hardcoding `Accept` in the rest layer.

https://github.com/pulumi/pulumi-pulumiservice/blob/48f156135cddc92865da748c6abad61b69a6ef57/provider/pkg/rest/resource.go#L693-L697

lukebward:
### Code review

Found 2 issues:

1. CHANGELOG entry for `pulumiapi.Client.CreatePolicy` describes a test-only/internal helper that shouldn't be in the changelog (CLAUDE.md says "DO NOT add CHANGELOG entries for: Test-only changes (adding tests, fixing tests, test infrastructure)"). The entry itself states it is "internal API surface" used by integration tests.

https://github.com/pulumi/pulumi-pulumiservice/blob/4cf85cc63ee011d45208f26c0efa860053a4c519/CHANGELOG.md#L57-L61

2. `examples/hammer_v2_test.go` is tagged `//go:build hammerv2` only, but references `v2Cases`, `runV2Case`, and `sweepV2TestState`, which are defined under `//go:build v2` in `examples_v2_test.go` and `sweep_v2_test.go`. The documented `go test -tags hammerv2 …` command in the file header fails to compile (`undefined: sweepV2TestState`). Either widen the v2/sweep files' build tag to `v2 || hammerv2`, or change this file's tag (and the doc command) to require both.

https://github.com/pulumi/pulumi-pulumiservice/blob/4cf85cc63ee011d45208f26c0efa860053a4c519/examples/hammer_v2_test.go#L1-L5

examples/v2/access-tokens/typescript/package.json
```diff
@@ -0,0 +1,11 @@
+{
+	"name": "pulumicloud-v2-access-tokens",
+	"devDependencies": {
+		"@types/node": "^18"
```
lblackstone: Is this right?

lblackstone (approved):

lblackstone:
I'm not sure if you're already tracking this, but we'll want to add a dedicated migration guide that's agent-friendly. Can be a follow up.

infra/test-aws-oidc/README.md
iwahbe, line 1: Why delete this?

provider/pkg/apiclient/apiclient.go
```diff
@@ -16,6 +16,10 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
+// AcceptMediaType is the media type sent on the Accept header for Pulumi
+// Cloud API requests; the +N suffix is the wire-protocol version.
+const AcceptMediaType = "application/vnd.pulumi+8"
```
iwahbe: I think this should be 9 now.

provider/pkg/config/config.go
```diff
@@ -105,6 +109,9 @@ func (c *Config) Configure(context.Context) error {
 	if c.APIURL == "" {
 		c.APIURL = os.Getenv(EnvVarPulumiBackendURL)
 	}
+	if c.APIURL == "" {
+		c.APIURL = os.Getenv(EnvVarPulumiAPI)
+	}
```
iwahbe: Neither of these are necessery in `infer` land, the `SetDefault` call handles both of these.

provider/pkg/rest/metadata.go
```diff
@@ -0,0 +1,127 @@
...
+	"encoding/json"
+	"fmt"
+)
+
+// Metadata is the parsed contents of metadata.json: the Pulumi-only overrides
+// (operationId → resource mapping, forceNew, secret, aliases, etc.) layered
+// over the OpenAPI spec.
+type Metadata struct {
+	// Version pins the schema version of this file. Currently 1.
+	Version int `json:"version"`
```
iwahbe: Why do you need this? Do we ever intend to version the metadata file separately from the provider? Isn't the metadata file just for the provider?

examples/v2/access-tokens/typescript/package.json
lukebward: I am actually going to bump to ^22.  I think v1 is on 14/16

provider/pkg/rest/metadata.go
```diff
@@ -0,0 +1,127 @@
...
+	// operationId.
+	RequireImport bool `json:"requireImport,omitempty"`
+
+	// Description overrides the generated resource description; empty
+	// falls back to the create op's description.
+	Description string `json:"description,omitempty"`
+
+	// Examples are PCL snippets rendered as `## Example Usage` blocks.
+	// SDK codegen runs `pulumi convert` per target language at gen time.
+	Examples []string `json:"examples,omitempty"`
```
iwahbe, line 86: I love this!

lukebward:
> I'm not sure if you're already tracking this, but we'll want to add a dedicated migration guide that's agent-friendly. Can be a follow up.

I will make a quick follow up PR for adding this doc in.

provider/pkg/rest/resource.go
```diff
@@ -0,0 +1,959 @@
...
+	for k, v := range a.AllStable {
+		bv, ok := b.GetOk(k)
+		if !ok {
+			return false
+		}
+		if !v.Equals(bv) {
+			return false
+		}
+	}
+	return true
```
iwahbe: This could  be simplified  to `return property.New(a).Equals(property.New(b))`

provider/tools/openapi-fetch/main.go
iwahbe, line 1: Any reason why this whole file can't  be replaced with `curl "${PULUMI_CLOUD_OPENAPI_URL:-https://api.pulumi.com/api/openapi/pulumi-spec.json}" | jq --sort-keys`?

provider/pkg/rest/metadata.go
lukebward: I would love this too, but sadly this comment is out of date.  I see that you added that fix in, so I will swap back to PCL in a follow up PR. https://github.com/pulumi/pulumi/issues/16384

iwahbe: Until the next codegen bug.

infra/test-aws-oidc/README.md
lukebward: Whoops, caught in the claude-fire :/

provider/pkg/rest/metadata.go
lukebward: https://github.com/pulumi/pulumi-service/issues/43504#issue-4481314404 I will follow up on this though

provider/pkg/rest/resource.go
lukebward: Ugh.  Thank you...

provider/pkg/config/config.go
lukebward: If I switch to `SetDefault`, `pulumi import` doesn't seem to persist the properly and falls back to prod API.

iwahbe: Hmm, can you file a bug in pulumi-go-provider for that?

lukebward: https://github.com/pulumi/pulumi-go-provider/issues/527

lukebward:
### Code review

Found 4 issues:

1. `Configure` mutates the per-provider transport ref and the global `rest.SetTransportResolver` *before* checking the error from `pulumiapi.NewClient`. If `NewClient` fails (e.g. unparseable URL), the global resolver is left pointing at the just-built transport before the error returns, so other code paths see a poisoned resolver.

https://github.com/pulumi/pulumi-pulumiservice/blob/2548b8c4e428728c779f44b49baf61098585f1bb/provider/pkg/provider/provider.go#L369-L398

2. `MergeSecretValue` compares `cipherValue.Value == oldCipherValue.Value`, but the new `apitype.SecretValue` (`provider/pkg/apitype/secret_value.go`) stores ciphertext in `.Ciphertext` and leaves `.Value == ""` for ciphertext-only API responses (see `UnmarshalJSON`). For refresh of an existing secret, both sides are `""`, the comparison is always true, and a Cloud-side rotation is never detected — the old plaintext is preserved in state.

https://github.com/pulumi/pulumi-pulumiservice/blob/2548b8c4e428728c779f44b49baf61098585f1bb/provider/pkg/util/secret.go#L138-L148

3. `ListOrgMembers` docstring promises "a logged warning" on a frontend roster failure, but the code silently sets `frontend = nil` with no log call — `frontendErr` is discarded. A 403/timeout on the frontend roster will produce a truncated member list with no diagnostic.

https://github.com/pulumi/pulumi-pulumiservice/blob/2548b8c4e428728c779f44b49baf61098585f1bb/provider/pkg/pulumiapi/members.go#L170-L178

4. Package doc in `provider/pkg/cloud/spec.go` still references the "v2 resource surface" and "v2 CRUD dispatch", but the namespace was renamed v2→v1 in 000078ff (and the same package now wires `withCloudV1Schema`). The comment is stale.

https://github.com/pulumi/pulumi-pulumiservice/blob/2548b8c4e428728c779f44b49baf61098585f1bb/provider/pkg/cloud/spec.go#L14-L22

provider/tools/scaffold-metadata/main.go
```diff
@@ -0,0 +1,1306 @@
...
+					"  INFO: token pinned at %q; heuristic now suggests %q (preserving pin)\n",
+					existing, d.Token)
+			}
+		} else {
+			entry["token"] = d.Token
+		}
+	}
+
+	if d.IDFormat != "" {
+		entry["idFormat"] = d.IDFormat
```
fallimic: Not sure if we should have the same `if existing, has` check here. We will overwrite hand edits of IdFormat

provider/pkg/rest/check_test.go
```diff
@@ -0,0 +1,315 @@
...
+			IDFormat:            idFormat,
+			DeleteBeforeReplace: dbr,
+		},
+	}
+}
+
+func TestCheckNormalizesEnumCase(t *testing.T) {
+	spec := syntheticSpec(t)
+	r := fooResource(spec, nil, "", false)
+	resp, err := r.Check(context.Background(), p.CheckRequest{
```
fallimic: Throughout tests we should prefer t.Context() instead of context.Background()

provider/pkg/rest/resource.go
```diff
@@ -0,0 +1,947 @@
...
+	} else if ok {
+		state = fetched
+	} else {
+		state = mergeMaps(state, req.State)
+	}
+	return p.UpdateResponse{Properties: state}, nil
+}
+
+// Delete fires the delete op (if declared). Without one the engine drops
+// state silently. Path params come from req.OldInputs with req.Properties
```
fallimic: This comment is opposite of what we have below in `mergeMaps(req.Properties, req.OldInputs)`

fallimic (approved):

PR #815, opened 2026-05-08
Drop net6.0 in favour of net8.0 for the .NET SDK

iwahbe:
## Summary

- Pulumi NuGet 3.106.0 was published at 2026-05-08T10:31 UTC and dropped \`net6.0\` support.
- Pulumi codegen still emits \`<TargetFramework>net6.0</TargetFramework>\` into the generated \`Pulumi.PulumiService.csproj\`, so every regenerated dotnet SDK fails with NU1202 (\"Package Pulumi 3.106.0 is not compatible with net6.0\").
- This bumps the SDK to \`net8.0\` and adds a Makefile post-process step so future regenerations stay on \`net8.0\` until pulumi codegen catches up. Example projects already target \`net8.0\`, so no other source changes are required.

## Test plan

- [x] \`make generate_dotnet\` produces a csproj with \`net8.0\` (verified locally)
- [x] \`make build_dotnet\` succeeds locally against \`Pulumi 3.106.0\`
- [ ] CI \`build_sdk (dotnet)\` passes (this PR)

## Notes

The Makefile is autogenerated by ci-mgmt. The \`perl -i -pe\` step added here will be wiped out the next time ci-mgmt regenerates the file unless ci-mgmt is updated in parallel. ci-mgmt already has an \`update_to_dotnet8\` migration but only runs it for \`bridged-provider\` / \`external-bridged-provider\` templates; this provider uses the \`generic\` template. A follow-up to widen that migration (or to bump the codegen template itself) is the proper long-term fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Breaking Changes
+- The .NET SDK (`Pulumi.PulumiService`) now targets `net8.0` and no longer supports `net6.0`. Pulumi NuGet 3.106.0 dropped `net6.0`, which made the previously-targeted framework unbuildable. Consumers must build against the .NET 8 SDK.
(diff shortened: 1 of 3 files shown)


iwahbe:
Not needed because of https://github.com/pulumi/pulumi-dotnet/pull/997

PR #836, opened 2026-05-14
Migrate `Environment` to `infer`

iwahbe:
## Summary
- Migrates the `Environment` resource to the `infer` framework. The legacy `PulumiServiceEnvironmentResource` (manual gRPC) is removed in favour of an `infer.Resource`-shaped `Environment` with `EnvironmentInput` / `EnvironmentState` structs.
- The Environment resource needs both the Pulumi Cloud client (for the metadata UUID) and the upstream Pulumi ESC HTTP client (for the YAML body); it reaches the latter via the existing `config.GetEscClient`, which the provider already vends alongside `config.GetClient`.
- Removes the manual `pulumiservice:index:Environment` block from `manual-schema.json`. Infer regenerates the schema with the same wire shape; the only diff is the addition of `replaceOnChanges: true` annotations on `organization`, `project`, and `name`, which were previously enforced in the legacy resource's custom `Diff`.

## Test plan
- [x] `make provider`
- [x] `make build_sdks`
- [x] `go test -short ./...` in `provider/pkg`
- [x] `make lint`
- [ ] CI integration tests

## Notes
- `yaml` is typed as `types.AssetOrArchive` to preserve the published schema (`pulumi.json#/Asset`).
- `EnvironmentState` deliberately does not embed `EnvironmentInput`: keeping `project` required in state (with a default) matches the legacy SDK's non-nullable `Project` output even though `project` is optional on input.


iwahbe:
[Metabase](https://metabase.corp.pulumi.com/question/7683-inputs-outputs-for-a-resource-duplicate?RESOURCE_TYPE=pulumiservice%3Aindex%3AEnvironment) shows we don't need an `__inputs` migration.

PR #837, opened 2026-05-14
Migrate `PolicyGroup` to `infer`

iwahbe:
## Summary
- Migrates the `PolicyGroup` resource to the `infer` framework, replacing
  the legacy gRPC `PulumiServicePolicyGroupResource` implementation.
- Removes the manual `PolicyGroup`, `PolicyGroupStackReference`,
  `PolicyGroupPolicyPackReference`, and `PolicyGroupPolicyPackReferenceInput`
  schema blocks from `manual-schema.json`; infer regenerates them from the
  Go struct definitions.
- Preserves the legacy resource semantics: order-independent diff on
  stacks/accounts/policyPacks, child-account skip-on-remove when the parent
  is still in the input set, `version`-strip in Check so legacy programs
  that still set the server-derived field upgrade cleanly, and
  replaceOnChanges for name / organizationName / entityType / mode.
- `Update` sends each batch op in its own request to dodge the Cloud's
  reorder/upsert-by-name behavior, and recovers the real state for the
  checkpoint if a mid-batch op fails; policy-pack equality compares config,
  treating a nil and an empty config map as equal.

## Notes
- The generated SDK type for the policy pack's `version` field changes from
  `number` to `integer`, matching the API client's `int` field. This is a
  minor SDK-shape change but more accurate.
- Existing tests for diff order-independence, policy-pack equality (including
  config), the parent-account child-skip rule, and the `version`-strip
  behavior cover the new resource shape.

## Test plan
- [x] `make provider`
- [x] `make build_sdks`
- [x] `go test ./...` in `provider/pkg`
- [x] `make lint` (provider + sdk + examples)
- [ ] CI integration tests


iwahbe:
[Metabase](https://metabase.corp.pulumi.com/question/7683-inputs-outputs-for-a-resource-duplicate?RESOURCE_TYPE=pulumiservice%3Aindex%3APolicyGroup) shows that an `__inputs` migration isn't necessary.

PR #875, opened 2026-05-27
Add PolicyPack resource

lukebward:
## Summary
- Adds `PolicyPack` resource for publishing policy packs to Pulumi Cloud directly from a Pulumi program.
- Source directory is tarballed and uploaded on Create; policy metadata is auto-extracted by running the pack's language analyzer plugin (matching `pulumi policy publish` behavior) and may be overridden inline.

## Test plan
- [ ] CI passes
- [ ] `ts-policy-pack` and `yaml-policy-pack` example tests pass


aacotroneo (commented):
Did a structural pass with Claude and I'm missing a few things.

The headline finding is that this PR reimplements packaging logic that already exists in `pulumi/pulumi`, but it diverges quite a bit.

## How `pulumi policy publish` actually works

The canonical entry point is [`cloudPolicyPack.Publish`](https://github.com/pulumi/pulumi/blob/d1eac119328d18fa01c4902ba0a655f3feae1ac9/pkg/backend/httpstate/policypack.go#L397-L461). It does three things: introspect, tarball, upload. The tarball step is [a runtime switch](https://github.com/pulumi/pulumi/blob/d1eac119328d18fa01c4902ba0a655f3feae1ac9/pkg/backend/httpstate/policypack.go#L430-L444):

- **nodejs** → [`npm.Pack`](https://github.com/pulumi/pulumi/blob/d1eac119328d18fa01c4902ba0a655f3feae1ac9/sdk/nodejs/npm/manager.go#L147) — shells out to the user's package manager, respects `.npmignore` / `package.json:files`, handles lockfiles, wraps in `package/`.
- **everything else** → [`archive.TGZ(pwd, "package", true)`](https://github.com/pulumi/pulumi/blob/d1eac119328d18fa01c4902ba0a655f3feae1ac9/sdk/go/common/util/archive/archive.go#L35) — honors `.gitignore` (incl. parent `.gitignore`s when at the root), wraps in `package/`.

The install side ([`installRequiredPolicy`](https://github.com/pulumi/pulumi/blob/d1eac119328d18fa01c4902ba0a655f3feae1ac9/pkg/backend/httpstate/policypack.go#L499-L577)) extracts the tgz and then runs `os.Rename(tempDir/package, finalDir)` — **the `package/` prefix is load-bearing on the consumer side.**

The upload side ([`client.PublishPolicyPack`](https://github.com/pulumi/pulumi/blob/d1eac119328d18fa01c4902ba0a655f3feae1ac9/pkg/backend/httpstate/client/client.go#L1737-L1836)) also validates the version tag via [`validatePolicyPackVersion`](https://github.com/pulumi/pulumi/blob/d1eac119328d18fa01c4902ba0a655f3feae1ac9/pkg/backend/httpstate/client/client.go#L1874-L1886) (regex `^[a-zA-Z0-9-_.]{1,100}$`) and PUTs through the retry-enabled `restClient.HTTPClient()`.

## Deltas vs this PR

| Concern | `pulumi/pulumi` | This PR |
|---|---|---|
| nodejs packing | `npm.Pack` (respects `.npmignore`, `files`, lockfile) | Bespoke `WalkDir` with hardcoded skip list |
| other runtimes | `archive.TGZ` honors `.gitignore` (and parent `.gitignore`s) | Same hardcoded list, no `.gitignore` |
| tarball entry layout | **everything under `package/`** | files at the tar root |
| version source | `analyzerInfo.Version` (from `package.json` / `PulumiPolicy.yaml`) | user-supplied `versionTag` |
| version validation | regex check, clear error | none; bad chars → opaque 400 |
| signed-URL PUT | retries via `restClient.HTTPClient().Do(..., retryAllMethods)` | raw `httpClient.Do`, no retries |
| Policy fields passed through | full `apitype.Policy` incl. `Severity`/`Framework`/`Tags`/`RemediationSteps`/`URL` | dropped |
| ConfigSchema type | typed `apitype.PolicyConfigSchema` with `map[string]*json.RawMessage` | `map[string]any` + ad-hoc `normalizeConfigSchema` |

## What I'd recommend

**1. Probable correctness bug — missing `package/` prefix.** `tarballDirectory` writes entries at the tar root, but the canonical extractor expects them under `package/`. If the Cloud's policy-execution sandbox uses the same extraction logic (very likely — it's the canonical path), packs published by this resource won't run.  ideally the layout should match. Do these packs work? (That's the final test!);

** Not sure how much we mean "matching `pulumi policy publish` behavior." The analyzer-introspection part matches. The packaging does not — no `npm pack`, no `.gitignore`/`.npmignore`, no `package/` prefix, no version validation, no PUT retries. 

**3. Code-judo: delete `policy_pack_archive.go` and call the canonical helpers.** The Pulumi SDK is already a dependency (we use it for `plugin.PolicyAnalyzer`). Drop in:

```go
// nodejs
packTarball, err := npm.Pack(ctx, npm.AutoPackageManager, in.SourcePath, io.Discard)

// everything else  
packTarball, err := archive.TGZ(in.SourcePath, "package", true /*useDefaultExcludes*/)
```

That deletes `policy_pack_archive.go` (~117 LOC) and `policy_pack_archive_test.go` (~146 LOC), fixes the layout/.gitignore/.npmignore/lockfile issues for free, and makes the CHANGELOG claim actually true.

## Other structural items worth a skim

These are lower-priority than the packaging issue but worth addressing while we're here:

- **`CustomDiff` is currently re-implementing `replaceOnChanges`.** The struct already tags Organization/Name/DisplayName/VersionTag with `provider:"replaceOnChanges"`. Implementing `CustomDiff` overrides infer's default — so the `Diff` body manually re-enumerates the same fields. If we lean on the framework, `Diff` shrinks a lot (or disappears, depending on how we handle drift).

- **The `policy_group.go` changes are a stowaway.** Serializing batched group updates one-at-a-time to dodge a backend reorder bug is unrelated to PolicyPack, lacks a test for the new partial-error path (existing tests were only changed `=` → `append` to not break), and arguably belongs in `pulumiapi.BatchUpdatePolicyGroup` rather than in the resource handler. The comment on the Create path is also explicit that it degrades a working batched call to match Update's brokenness "for consistency" — that's spreading the problem, not solving it. I'd split this into its own PR.

- **Three near-identical Policy structs + four trivial helpers.** `PolicyPackPolicyInput` (resource), `pulumiapi.Policy` (API), and `plugin.AnalyzerPolicy` (SDK) all get copied between each other via `toAPIPolicies` / `convertAnalyzerConfigSchema`, with `normalizeConfigSchema` papering over a backend quirk and `policiesEqual` being a one-line `reflect.DeepEqual` wrapper. Aligning `PolicyPackPolicyInput` with `apitype.Policy` (or `pulumiapi.Policy`) would let most of this disappear — and would also un-drop `Severity`/`Framework`/`Tags`/`RemediationSteps`/`URL` for inline policies.

- **`PublishPolicyPack` does three jobs in one function.** Metadata POST + signed-URL PUT (raw httpClient, bypassing `c.do`) + complete POST + deferred rollback. Extracting `uploadToSignedURL` would let the tests stop having to spin up two `httptest.Server`s and a multi-route dispatcher to exercise one error path.

lukebward:
@aacotroneo did you use the review skill that Joe sent?

aacotroneo (approved):

PR #881, opened 2026-05-29
ci: restore api-release workflows under pulumiservice- prefix

lukebward:
ci-mgmt regeneration deletes custom workflow files unless they're prefixed with the provider name, which removed the `pulumiservice:api:*` release pipeline. This re-adds it, prefixed so future regenerations skip rather than delete it.

- Re-add `refresh-api-spec`, `tag-api-release`, and `api-tests` as `pulumiservice-*` files
- `api-tests` self-triggers on release tags instead of being called from the generated `release.yml` / `prerelease.yml`
- Fix the CONTRIBUTING link to the renamed tag workflow

CI + docs only — no CHANGELOG entry (apply `impact/no-changelog-required`). Note: api-tests no longer hard-gate `publish` (that job lived in the ci-mgmt-generated release files); they run on every PR and on the release tag. Add `api-tests` as a required status check on `main` to re-gate.


Makefile
```diff
@@ -206,7 +206,10 @@ install_dotnet_sdk: .make/install_dotnet_sdk
 	; fi
 	@touch $@
 install_go_sdk:
-install_java_sdk:
+install_java_sdk: .make/install_java_sdk
+.make/install_java_sdk: .make/build_java
+	cd sdk/java/ && gradle --console=plain publishToMavenLocal
+	@touch $@
```
pose: Sorry @lukebward, the next `ci-mgmt` will override the `Makefile` since this file is generated. This is required by the API gen, right?

pose (changes requested):

Makefile
pose: (Asking to find a way of giving you an escape hatch from `ci-mgmt` point-of-view)

lukebward: Yes, this is required.  https://github.com/pulumi/pulumi-pulumiservice/pull/878/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

pose: Fix coming: https://github.com/pulumi/ci-mgmt/pull/2284

pose: I need to design/think a bit around this other issue: https://github.com/pulumi/ci-mgmt/issues/2285 . @lukebward if you are blocked, feel free to commit your changes and I'll disable the re-gen until we have a path forward.

lukebward: I am not blocked at the moment as I am not doing active development here, but do let me know when progress is made.  Thanks Alberto!

pose: @lukebward Fix for additional dependencies required: https://github.com/pulumi/pulumi-pulumiservice/pull/885

pose: Both fixes have now landed.

pose (approved):

Makefile
pose: These should now be removed as these is already done at a difference place

pose (changes requested):
After removing Makefile changes, we should be good to merge 🥳

lukebward:
> After removing Makefile changes, we should be good to merge 🥳

did you add the fix elsewhere?  These makefile changes are necessary for the tests to pass here

pose:
Let me have a look at the test and get back to you!

pose:
Done, kept the test. Moved the publish out of the Makefile into a regen-safe `.mk/sdk-install.mk`: `install_java_sdk` is generated empty, so it just adds a prerequisite plus a recipe on a new `.make/` target, nothing overrides the generated Makefile, so `make ci-mgmt` won't clobber it. Also pointed the api-tests java lane at `make build_java install_java_sdk`. Makefile is now untouched vs `main`.

pose (approved):

PR #937, opened 2026-07-02
Extract repeated literals to constants and annotate gosec/govet false positives for golangci-lint 2.12.2

pose:
The ci-mgmt workflow update bumps golangci-lint to 2.12.2, which enables stricter `goconst`/`gosec`/`govet` checks. This hoists repeated string literals into named constants (predominantly in tests), replaces the deprecated `reflect.Ptr` with `reflect.Pointer`, and adds line-scoped `//nolint` annotations (with reasons) for the `gosec` G101 test fixtures, the G704/G122 false positives (request host pinned to the operator-configured API base URL; hashing the caller's own local policy-pack dir), and the `govet` inline-analyzer noise on generic `slices.Sort` calls. Verified locally with golangci-lint 2.12.2 (uncapped) reporting 0 issues; unit tests pass.

Part of #936


guineveresaenger (approved):
Needs the Node version bump but otherwise LGTM

provider/pkg/pulumiapi/goconst_test.go
```diff
@@ -0,0 +1,43 @@
+package pulumiapi
```
pose: Do no do this. Add to the required test files each constant.

provider/pkg/pulumiapi/goconst_consts.go
```diff
@@ -0,0 +1,12 @@
+package pulumiapi
+
+// Constants extracted to satisfy the goconst linter for repeated string
```
pose: Do no do this. Add to the required test files each constant.

provider/pkg/pulumiapi/goconst_consts.go
```diff
@@ -0,0 +1,12 @@
+package pulumiapi
```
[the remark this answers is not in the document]
pose: Resolved in 35e452a. These aggregate `goconst_*.go` files were removed entirely and each constant now lives in the existing file that uses it, so there are no new files requiring copyright headers.

provider/pkg/pulumiapi/goconst_test.go
pose: Done in 35e452a. Each constant is now declared in the file that uses it (per-usage placement), and the separate `goconst_*.go` / `consts_test.go` aggregate files are deleted.

provider/pkg/pulumiapi/goconst_consts.go
pose: Done in 35e452a. Each constant now lives in the source file that uses it, and this aggregate file is removed.

pose:
Thanks @guineveresaenger! The Node bump is a separate, pre-existing CI issue: `@pulumi/pulumi@3.250.0` requires Node >=22 while main's workflow still pins Node 20, so `build_sdk`/`api-tests` fail on any PR to main until the ci-mgmt workflow bump lands. This PR's `lint` job is green.

PR #1014, opened 2026-07-27
Expose policy pack publisher on getPolicyPacks/getPolicyPack

### LINKED ISSUE #1013: getPolicyPacks should expose the policy pack publisher
`getPolicyPacks` returns only `name`, `displayName`, `versions`, and `versionTags`. There is no publisher or ownership field, so a program cannot tell a Pulumi-published pack (`cis-aws`, `pulumi-best-practices-aws`, …) from one the org published itself.

That filter is the common case for anything that manages Policy Groups programmatically — "apply Pulumi's compliance packs, leave our own alone" currently has to be written as a name-matching heuristic against Pulumi's `<framework>-<cloud>` convention, which silently goes stale whenever Pulumi publishes a pack that doesn't fit the pattern.

`getPolicyPacks` (and `getPolicyPack`, for consistency) should expose the publisher so downstream clients can filter on it.

### This does not need an API change

`GET /api/orgs/{org}/policypacks`, which `getPolicyPacks` wraps, does not return the field. But `GET /api/registry/policypacks?orgLogin={org}` does:

```json
{"source": "pulumi",  "publisher": "pulumi", "name": "cis-aws",       "version": "1.0.2"}
{"source": "private", "publisher": "pequod", "name": "demo-policies", "version": "0.0.5"}
```

Pulumi-published packs are `source: "pulumi", publisher: "pulumi"`; an org's own are `source: "private", publisher: "<org>"`.

So the provider can populate the field by calling the registry list endpoint once per invoke and joining on pack name — which is what the Pulumi Cloud console already does to render its publisher column and its policy group detail view. Two options:

1. Add `source` and `publisher` to the existing `getPolicyPacks` result, backfilled from that join. Non-breaking, and callers get the filter for free.
2. Add a separate `getRegistryPolicyPacks` invoke wrapping the registry endpoint directly, exposing `source`, `publisher`, `accessLevel`, and `enforcementLevels`. More faithful to the API, but leaves callers to do the join.

(1) seems the better fit for the common case, and doesn't preclude (2) later.

Two caveats for whoever picks this up. The registry list handler is named `ListPolicyPacksDeprecated` (a newer POST form takes a request body), so it's worth confirming which form to build on. And the join key is the bare pack name, while the registry namespaces packs as `{source}/{publisher}/{name}` — if two publishers ever ship a pack with the same name, a name-keyed join is ambiguous, and the org endpoint has no way to disambiguate. Adding `source`/`publisher` to the org endpoint itself would be the durable fix; that's a service-side change and is not required for this issue.

jkodroff:
Fixes #1013

`getPolicyPacks` returned only `name`/`displayName`/`versions`/`versionTags`, so a program couldn't tell a Pulumi-published pack (`cis-aws`) from one the org published itself. "Apply Pulumi's compliance packs, leave ours alone" had to be a name-matching heuristic against Pulumi's `<framework>-<cloud>` convention, which goes stale whenever Pulumi ships a pack that doesn't fit the pattern.

Both invokes now return `source` and `publisher`, backfilled from the registry. No service-side change needed.

## Two findings worth reviewing first

**1. A named schema type is load-bearing, not cleanup.** Pulumi's schema format only supports `$ref` in array `items` — the previous inline object was flattened to a bare `{"type": "object"}` during generation, so every SDK rendered the result as an untyped string map. Adding `source`/`publisher` inline would have reached YAML programs only and changed nothing in any SDK. Hence `PolicyPackSummary`, and hence the breaking-change entry.

**2. The Go SDK wasn't merely mistyped — it was returning nothing.** `[]map[string]string` silently drops any pack object holding a non-string value. The provider always emits `versions` as a list of integers, so *every* pack was dropped and `getPolicyPacks` returned an **empty list with no error**. Verified against the real unmarshal path with the Go SDK's mock harness. Go programs that appeared to work were reading nothing, which makes this side of the change a fix rather than a regression.

## API routes

Per the vendored spec (`provider/pkg/cloud/spec.json`):

| Route | `deprecated` | `x-pulumi-route-property` |
|---|---|---|
| `GET /api/preview/registry/policypacks` | **`true`** | `{"Deprecated": true, "SupersededBy": "ListPolicyPacks_preview_registry_post", "Visibility": "Preview"}` |
| `POST /api/preview/registry/policypacks` (**used**) | `false` | `{"Visibility": "Preview"}` |
| `GET /api/orgs/{org}/registry/policypacks/{name}` (**used**) | `false` | `{}` |

The issue proposed the GET; the spec explicitly supersedes it with the POST at the same path, so this builds on the POST. The single-pack case uses the org-scoped route, which has no deprecation or preview marker and resolves the name in the org's own namespace — no client-side ambiguity to handle.

`Visibility: Preview` isn't exotic here (147 of 603 operations), and the provider already ships a GA resource on preview routes — all of `InsightsAccount` runs on `/api/preview/insights/...`.

## Failure behavior

- **404/405 on the registry route** → backend doesn't serve it (older or self-hosted Cloud). Return packs unannotated + warn. `getPolicyPacks` worked on those backends before this change and must keep working.
- **Everything else (403/500/network/parse)** → fail the invoke. A silently unannotated result makes a `publisher == "pulumi"` filter match nothing, attach zero policy packs, and still exit 0 — compliance coverage drops with nothing in the diff.
- **Pack absent from the registry** → degrades individually, not the whole call.
- **Name claimed by two publishers** → resolves to nothing. Mis-attribution is the same failure this feature exists to prevent, and first-wins would be nondeterministic since registry list order is unspecified.

Fields are **omitted**, never `""`, so they generate as `*string` / `string | undefined` / `Optional[str]` and a degraded result stays distinguishable from real data.

`getPolicyPack` deliberately soft-fails where `getPolicyPacks` hard-fails — the caller already named the pack, so the empty-filter hazard doesn't exist. Called out in a code comment so it doesn't get "fixed" to match.

`ListPolicyPacks` is left untouched: it backs the `PolicyPack` resource's Read path, which has no use for registry metadata and shouldn't fan out to a second endpoint on every refresh.

## Tests

Written TDD — client tests confirmed RED before implementation.

- **20 API client subtests** (`registrypolicypacks_test.go`), including 403/500-fail, 404/405-degrade, collision, order-independence, and org-list-failure-skips-registry.
- **9 provider-level tests** (`policypacks_invoke_test.go`) over a surface that previously had **zero** coverage. These assert on the unmarshalled property map, proving the fields cross the RPC boundary — something no client-level test can show. Includes asserting field *absence* (not `""`) on the degraded path.
- **Schema regression test** pinning the `$ref`, so reverting to an inline items object fails loudly instead of silently dropping the fields from SDKs.
- **Integration coverage** added to the existing `yaml-policy-groups` example rather than a new directory, per CLAUDE.md.

Provider lint 0 issues; examples lint matches its pre-existing baseline exactly.

## Still needs a real org before this leaves draft

The registry list's `access` filter defaults to `enabled`, and the spec's enum comments are ambiguous. If that default excludes added-public packs, `cis-aws` comes back unannotated and the feature misses its headline case. Unit tests prove we degrade gracefully; only a real run proves we don't degrade *universally*:

```
cd examples && go test -v -run TestYamlPolicyGroupsExample -tags yaml -timeout 10m
```

If Pulumi-published packs come back without a publisher, the fix is passing `access` explicitly — the degradation path already handles it without erroring.

## Breaking change

`getPolicyPacks.policyPacks` becomes typed `PolicyPackSummary[]`. Wire format unchanged; invoke results aren't persisted, so no state migration, refresh drift, or import impact. Per-language migration is in the CHANGELOG — notably Python, where the generated class is a `dict` subclass keyed by snake_case, so `pack["displayName"]` now raises `KeyError` (`pack.display_name` works).


lukebward:
Can we get CI to pass?

jkodroff:
@lukebward All checks pass!

lukebward:
It looks like we recently changed some Policy Pack routes, I am going to merge in a separate PR that generates the new  `spec.json`.  Once that's in I'll let you know if and what changes need to be made here, and then we can merge it in

lukebward:
Should be good.  Just rebase and send it sir!

PR #1036, opened 2026-08-13
Support private registry credentials for executorImage

### LINKED ISSUE #170: DeploymentSettings.executorContext.executorImage should support passing in private image credentials

jkodroff:
Fixes #170.

`DeploymentSettings.executorContext` gains an optional `credentials` object (`username` plus a secret `password`), so a custom `executorImage` can be pulled from a private container registry. Pulumi Cloud has always accepted these credentials — they were simply not exposed by this provider (confirmed on the issue; it's in the API, just undocumented).

```yaml
executorContext:
  executorImage: "myreg.example.com/custom-pulumi:latest"
  credentials:
    username: "registry-user"
    password: ${imagePassword}
```

## Why `credentials` is a sibling, not nested

The Cloud API nests credentials inside the image (`{reference, credentials}`), but this provider types `executorImage` as a **required plain string**. Matching the API exactly would be a breaking change for every existing program and would churn all five SDKs, so `credentials` sits next to `executorImage` and the provider maps it into `Executor.ExecutorImage.Credentials`.

Nothing changes for programs that don't set it. `DockerImage.MarshalJSON` only switches from a bare string to the object form once `Credentials` is non-nil, and there's a test pinning the bare-string case.

## Two things worth a reviewer's eye

**The password is sent as `SecretValue{Secret: true}`.** Pulumi Cloud's `encryptSettings` walks the settings struct and encrypts only the values flagged secret, so a password sent as a bare string would be persisted unencrypted server-side. `TestCreateDeploymentSettings` pins the exact wire format.

**`Check` promotes the password to secret.** A schema `"secret": true` on a property nested inside a type gets no codegen support — every site in `pulumi/pulumi` that reads the flag iterates a resource's own top-level properties, and `additionalSecretOutputs` can't name a nested path. The generated `DeploymentSettings` SDK contains zero `secret` handling in any language as a result. Without the `Check` promotion, a program passing a plain literal would have it recorded verbatim in the state file's `inputs`.

## Verification

Against a real stack on app.pulumi.com. `GET .../deployments/settings`:

```json
{
  "executorImage": {
    "reference": "myreg.example.com/custom-pulumi:latest",
    "credentials": {
      "username": "registry-user",
      "password": {
        "secret": "[secret]",
        "ciphertext": "AAABAPheAtVINmjnSCSV9a3FC5RE3cZE7PJMiE/hF7Nx0zoaEEOKVp726XHArWIkX5QZU6rKDBA="
      }
    }
  }
}
```

Credentials land, and the password comes back encrypted — plaintext is unretrievable by design (`RedactSettings`). Note `executorImage` correctly serialized as an object, which only happens once credentials are present.

- **Refresh:** `refresh` → `2 unchanged`, `preview` after → `2 unchanged`. No perpetual diff, so the twin-value secret merge is correct against the live service.
- **State:** with the password hardcoded as a plain literal in the program, the exported state holds it as ciphertext under Pulumi's secret sentinel, and grepping the whole export for the literal returns **0 occurrences**.

## Tests

- 5 resource tests: create round-trip, backwards compatibility with no credentials, refresh merge, refresh with no prior cipher state, and import placeholder.
- 2 `pulumiapi` wire-format tests: the object form with `{"secret": ...}`, and the unchanged bare-string form.
- Full provider suite green; `golangci-lint` 0 issues (the 11 findings in `examples/` predate this branch).

## Follow-up

While verifying this I found the pre-existing `sourceContext.git.gitAuth` fields have the same defect — schema-declared secret, nothing honoring it. Confirmed on a live stack: a hardcoded git password appears verbatim in the exported state. Filed as #1037, and left out of this PR to keep it in scope.


lukebward (approved):
lgtm!

Issue #1038, opened 2026-08-13
api/deployments:Settings makes no secrecy claim about git auth or registry credentials


jkodroff:
## Summary

[#1037](https://github.com/pulumi/pulumi-pulumiservice/issues/1037) fixed git auth credentials being written to state in plaintext for `pulumiservice:index:DeploymentSettings`. The same credentials are reachable through the generated api-modules surface, which has no secret handling at all.

`pulumiservice:api/deployments:Settings` types every nested context as `pulumi.json#/Any` in the generated schema (`provider/cmd/pulumi-resource-pulumiservice/schema.json`):

```json
"sourceContext":    { "$ref": "pulumi.json#/Any", "description": "The source context defining where the source code is located." },
"executorContext":  { "$ref": "pulumi.json#/Any", "description": "The executor context defining the execution environment." },
"operationContext": { "$ref": "pulumi.json#/Any", "description": "The operation context defining pre-run and post-run commands and environment variables." },
```

So the schema makes no secrecy claim about anything inside them, including:

- `sourceContext.git.gitAuth.sshAuth.sshPrivateKey`
- `sourceContext.git.gitAuth.sshAuth.password`
- `sourceContext.git.gitAuth.basicAuth.password`
- `executorContext.executorImage.credentials.password` (see [#170](https://github.com/pulumi/pulumi-pulumiservice/issues/170) / [#1036](https://github.com/pulumi/pulumi-pulumiservice/pull/1036))
- `operationContext.environmentVariables` secret values

`pulumiservice:api/deployments:ScheduledDeployment.request` is likewise `pulumi.json#/Any` and carries the same `CreateDeploymentRequest` shape, so it can hold the same credentials.

## Why the `Any` typing happens

`provider/pkg/cloud/spec.json` *does* define these properly — `DeploymentSettings.sourceContext` is a `$ref` to `#/components/schemas/SourceContext`. The api-modules schema generation flattens the `$ref` to `pulumi.json#/Any` rather than emitting a corresponding Pulumi type. That flattening is what erases any opportunity to mark nested properties.

## Why nothing else covers it

The generic api-module `Check` (`provider/pkg/rest/resource.go:116`) only normalizes enum case, sorts set-like arrays, and generates autoNames. There is no secret handling anywhere in `provider/pkg/rest/` or `provider/pkg/cloud/` — `grep -rn "MakeSecret\|IsSecret\|KeepSecrets"` over both returns nothing.

[...]

PR #1039, opened 2026-08-13
Store DeploymentSettings git auth credentials as secrets

### LINKED ISSUE #1037: DeploymentSettings git auth secrets are written to state in plaintext when the program passes a non-secret value
## Summary

`pulumiservice:index:DeploymentSettings` declares four `sourceContext.git.gitAuth` properties as `"secret": true` in [`provider/pkg/provider/manual-schema.json`](https://github.com/pulumi/pulumi-pulumiservice/blob/main/provider/pkg/provider/manual-schema.json) (lines 78–116), but nothing enforces that at runtime. If a program supplies any of them as a plain (non-secret) value, the value is recorded verbatim in the stack's state file `inputs`.

**Three genuine secrets to fix:**

- `sourceContext.git.gitAuth.sshAuth.sshPrivateKey`
- `sourceContext.git.gitAuth.sshAuth.password`
- `sourceContext.git.gitAuth.basicAuth.password`

**One mislabel to correct:** `sourceContext.git.gitAuth.basicAuth.username` is declared secret but a git username is not sensitive. The fix there is to drop the `"secret": true` flag, not to start encrypting it — see below.

## Reproduction

Confirmed against a real stack on app.pulumi.com, not just in unit tests.

```yaml
gitAuth:
  basicAuth:
    username: "GITUSER-LITERAL-alice"
    password: "GITPASS-LITERAL-hunter2"
```

After `pulumi up`, `pulumi stack export` → `.inputs.sourceContext.git.gitAuth`:

```json
{
  "basicAuth": {
    "password": "GITPASS-LITERAL-hunter2",
    "username": "GITUSER-LITERAL-alice"
  }
}
```

Both literals are present verbatim in the exported state. The same is reached by `config.require("password")` — any value the engine does not see as secret, not just a hardcoded string.

For comparison, `executorContext.credentials.password` (fixed in #170) in the same state file:

```json
"password": {
  "ciphertext": "AAABABDDRDC+wT67Nnv3gJOC82vPQv0qf9+rbS95bcw3Y8Y+m0dWarxXWoJKUu5WMhbawSFDLQz3000="
}
```

## `basicAuth.username`: drop the flag rather than enforce it

A git username is not a secret, so the schema is simply wrong to mark it one. It should lose its `"secret": true` and be treated as an ordinary string.

Worth noting how differently it behaves from the other three today. `ToPropertyMap` writes it with a plain `resource.NewPropertyValue` (`provider/pkg/resources/deployment_settings.go:146-150`) instead of the `util` secret helpers the other fields use, so unlike the passwords it is never protected at any point. Measured before and after `pulumi refresh`:

| Field | inputs, after `up` | inputs, after `refresh` | outputs |
| --- | --- | --- | --- |
| `basicAuth.password` | plaintext | encrypted | `[secret]` |
| `basicAuth.username` | plaintext | still plaintext | plaintext |

Dropping the flag is a no-op in generated code — as established below, codegen ignores `secret` on properties nested inside types, so the flag has never produced any SDK behavior. It only ever set a false expectation for anyone reading the schema or the registry docs.

## Root cause

The schema's `"secret": true` is honored by SDK codegen, but **only for a resource's own top-level properties** — it is never applied to a property nested inside a `#/types/...` object.

In `pulumi/pulumi`, every SDK-generating site that reads the flag iterates `res.Properties` or `res.InputProperties` as a flat list, with no recursion into object types:

| Site | Effect |
| --- | --- |
| `pkg/codegen/nodejs/gen.go:824` | wraps a top-level secret **input** in `pulumi.secret(...)` |
| `pkg/codegen/nodejs/gen.go:925` | emits `additionalSecretOutputs` for secret **outputs** |
| `pkg/codegen/go/gen.go:2352,2390` | `secretProps` / `secretInputProps` |
| `pkg/codegen/python/gen.go:1564,1589` | `pulumi.Output.secret(...)` / `additionalSecretOutputs` |

The remaining `.Secret` references in `pkg/codegen` are schema binding and `gen_program.go` (`pulumi convert`), neither of which affects SDK behavior.

That codegen works as intended for top-level properties — in this provider, `api/agents:Pool` produces `additionalSecretOutputs: ["tokenValue"]`, and `api/esc:Environment` produces both `pulumi.secret(args.yaml)` on the input side and `additionalSecretOutputs` on the output side. By contrast the generated `DeploymentSettings` SDK contains **zero** occurrences of `secret` in any language, because all four of its secret-flagged properties are nested inside types.

For a nested secret the only remaining actor is the provider itself, at runtime. `DeploymentSettings` is a legacy raw-gRPC resource, so it gets no help from `pulumi-go-provider`'s `infer` machinery; its `Check` (`provider/pkg/resources/deployment_settings.go`) passes `req.News` through untouched apart from normalizing `operationContext.oidc.aws.duration`.

Corroborating detail: `examples/ts-deployment-settings/index.ts` passes the SSH password as `config.requireSecret("password")`. That would be unnecessary if the schema flag were sufficient.

## Scope

- **State `inputs`** — affected. Values the engine does not see as secret are written to state unencrypted.
- **State `outputs`** — not affected. `Create`/`Update` build outputs from the API response, which Pulumi Cloud redacts to `[secret]` plus ciphertext.
- **Pulumi Cloud** — not affected. The provider sends these as `SecretValue{Secret: true}`, so the service encrypts them at rest.
- **After a `pulumi refresh`** — the input is rewritten through `util.MergeSecretValue`, which wraps it in `resource.MakeSecret`. So the plaintext window is from the first `up` until a refresh, and reopens on any subsequent `up`.

Separately, `pulumiservice:api/deployments:Settings` types `sourceContext` as `pulumi.json#/Any`, so it makes no secrecy claim about these fields at all. Worth deciding whether that surface needs its own handling.

## Proposed fix

Two parts:

1. Promote `sshAuth.sshPrivateKey`, `sshAuth.password`, and `basicAuth.password` to secret in `Check`, the same way [#1036](https://github.com/pulumi/pulumi-pulumiservice/pull/1036) did for `executorContext.credentials.password`:

   ```go
   if credentials["password"].HasValue() && !credentials["password"].IsSecret() {
       credentials["password"] = resource.MakeSecret(credentials["password"])
   }
   ```

   With four call sites total it is worth extracting a small helper that takes a property path and applies this.

2. Remove `"secret": true` from `basicAuth.username` in `manual-schema.json` and regenerate the schema and SDKs.

## Backwards compatibility

The fix is not breaking, and existing stacks heal themselves on the next `up`:

- **No spurious diff.** `Diff` unmarshals both sides with `util.StandardUnmarshal` (`KeepSecrets: false`), so secrets are unwrapped before `olds.Diff(news)`. A change in secretness alone, with the value unchanged, produces `DIFF_NONE`.
- **State is rewritten even on a no-op.** `SameStep.Apply()` in `pulumi/pulumi` (`pkg/resource/deploy/step.go:173`) copies only `ID` and `Outputs` forward from the old state; `s.new.Inputs` — the newly checked inputs — is left intact, and `Res()` returns `s.new`, which is what gets snapshotted. So the plaintext is replaced by an encrypted secret without the user seeing any diff.
- **This transition already happens today.** `Read` puts the three password/key fields through `util.MergeSecretValue`, which wraps inputs in `resource.MakeSecret`. Verified on a live stack: after `pulumi refresh` the `basicAuth.password` literal is gone from the exported state and replaced with ciphertext, with the refresh itself reporting `2 unchanged`. The `Check` fix only makes `up` behave the way `refresh` already does.
- **Dropping the `username` flag changes nothing at runtime.** The field is plaintext everywhere today and stays that way; codegen never read the flag, so no SDK output changes either.

No state migration or version bump is needed.

## Migration note

Two things the fix cannot do, which are worth a CHANGELOG mention:

- A stack that is never `up`'d or refreshed again keeps its plaintext inputs.
- Prior state versions retained by the backend still contain the plaintext regardless of what the current version holds.

Anyone who has committed a state file, or used a backend where earlier versions are readable, should rotate the affected credentials.

## Notes

Found while implementing #170. That PR fixes the same class of bug for the executor image registry password it introduces; this issue tracks the four pre-existing fields, which were left alone to keep that PR in scope.

jkodroff:
Fixes #1037.

## The bug

`DeploymentSettings` declares four `sourceContext.git.gitAuth` credentials as `"secret": true` in `manual-schema.json`, but nothing enforced it at runtime. A program passing any of them as a plain value had the literal written verbatim into the stack's state `inputs`.

The schema flag only drives SDK codegen, and codegen reads it for a resource's own top-level properties. For a property nested inside a `#/types/...` object, **only the .NET generator acts on it** (it wraps the setter in `Output.CreateSecret`); Node.js, Python, Go and Java emit nothing. `DeploymentSettings` is a legacy raw-gRPC resource, so it gets no help from `infer` either — its `Check` passed `req.News` through untouched apart from normalizing the OIDC duration.

> Note: issue #1037 states codegen ignores the flag for nested properties in *all* languages. Regenerating the SDKs showed that's not true for .NET — C# programs were never exposed to this.

## The fix

- **`util.MakeNestedSecret(props, path...)`** — a path-based promoter that walks through secret-wrapped intermediates and no-ops on a missing or non-object level. There was no property-path helper in `util/` before; every resource hand-rolled `HasValue()` + `GetSecretOrObjectValue` chains.
- **`Check`** now promotes all four credentials via that helper.
- **`basicAuth.username` needed more than a `Check` change.** It was the one credential sent to Pulumi Cloud unflagged (`Secret: cascadeSecret || IsSecret()`), so the service stored it unencrypted and handed the plaintext straight back. Now that `Check` marks it secret, the service encrypts it and redacts it to `[secret]` on read — so the plain `NewPropertyValue` in `ToPropertyMap` would have written that marker into state as the username. It now goes through the same `Merge`/`Create`/`ImportSecretValue` path as the other three, and `toSourceContext` hardcodes `Secret: true` to match. This also drops the now-dead `cascadeSecret` tracking in `toSourceContext`, which had no remaining reader.

The schema is unchanged, so there is no SDK regeneration in this PR.

## Verification

Unit: `TestMakeNestedSecret` (7 cases), `TestDeploymentSettingsCheckGitAuthSecrets`, and create/refresh/import roundtrips for `basicAuth`. These are the first `Check` tests this resource has had. Confirmed non-vacuous — with the promotion disabled, `TestDeploymentSettingsCheckGitAuthSecrets` fails with `expected a secret, got {a-private-key}`.

Integration: `TestDeploymentSettingsExample` gains a post-`up` state assertion (via a new `runPulumiTestAfterUp` hook) that the plaintext `sshPrivateKey: "key"` the example already passes lands in state as a secret envelope.

Also verified end-to-end against app.pulumi.com with a `basicAuth` program:

| | before | after |
| --- | --- | --- |
| `username` in state `inputs` | `GITUSER-LITERAL-alice` | ciphertext |
| `password` in state `inputs` | `GITPASS-LITERAL-hunter2` | ciphertext |
| `refresh` / `preview` / re-`up` | — | `3 unchanged` |

And the upgrade path, by deploying with a provider built from `main` and then switching binaries: state inputs heal to ciphertext on the next `up`, reported as `3 unchanged`.

## Upgrade behavior

- State `inputs` heal on the next `pulumi up`, no diff shown.
- `sshPrivateKey` and both passwords were always sent flagged as secrets, so they were already encrypted at rest and their `outputs` already read `[secret]`.
- `basicAuth.username` is the exception worth calling out: because a change of secretness alone produces no diff, the value stays unencrypted server-side — and plaintext in state `outputs` — until the resource is next actually updated for some other reason. Verified: forcing an unrelated `repoDir` change flips the output to `[secret]`.

No state migration or version bump.

## Follow-up

#1038 covers the same class of gap on `pulumiservice:api/deployments:Settings`, which types `sourceContext` as `pulumi.json#/Any` and makes no secrecy claim at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,15 @@
 
 ### Bug Fixes
 
+- Fixed `DeploymentSettings` git auth credentials being written to state in plaintext when the program supplies them as non-secret values. All four `sourceContext.git.gitAuth` credentials (`sshAuth.sshPrivateKey`, `sshAuth.password`, `basicAuth.username` and `basicAuth.password`) are declared `secret` in the schema, but that flag only drives SDK codegen for a resource's own top-level properties and reaches a property nested inside a type only in the .NET SDK — so in every other language a plain value was recorded verbatim in the stack's state `inputs`. The provider now marks all four secret in `Check`, and `basicAuth.username` is handled as an encrypted twin-value secret on create, refresh and import like the other three. [#1037](https://github.com/pulumi/pulumi-pulumiservice/issues/1037)
+
+  Upgrade behavior for existing stacks:
+
+  - State `inputs` heal on the next `pulumi up`, with no diff shown.
+  - `sshPrivateKey` and both passwords were always sent to Pulumi Cloud flagged as secrets, so they were already encrypted at rest and their state `outputs` already read `[secret]`.
+  - `basicAuth.username` was previously sent unflagged, so Pulumi Cloud stored it unencrypted. Because the change of secretness alone produces no diff, the value stays unencrypted server-side — and plaintext in state `outputs` — until the `DeploymentSettings` resource is next actually updated for some other reason.
+
+  Two things this fix cannot undo: a stack that is never updated again keeps its plaintext inputs, and earlier state versions retained by the backend still contain the plaintext. **Anyone who has committed a state file, or uses a backend where prior state versions are readable, should rotate the affected credentials.**
 - Fixed `OidcIssuer` panicking during `pulumi preview` when its `url` input was an unknown value, such as another resource's output. The provider read the unknown input as a concrete string during `Check` and crashed; previews now complete and resolve `url` as computed. [#831](https://github.com/pulumi/pulumi-pulumiservice/pull/831)
 - Fixed `AgentPool.agentPoolId` output never being populated. The provider previously wrote the value under a state key that did not match the schema, so `pool.agentPoolId` was always undefined. Existing stacks pick up the correct value on the next `pulumi refresh`.
 - Fixed TeamEnvironmentPermission spurious replacement on upgrade from 0.29.2 caused by the optional `maxOpenDuration` field being serialized as an empty string in Check and Read [#751](https://github.com/pulumi/pulumi-pulumiservice/issues/751)
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -48,7 +48,17 @@ func TestDeploymentSettingsExample(t *testing.T) {
 	test.SetConfig(t, "my_secret", "my_secret_value")
 	test.SetConfig(t, "password", "my_password")
 	test.SetConfig(t, "imagePassword", "my_image_password")
-	runPulumiTest(t, test)
+
+	// The example passes sshPrivateKey as a plaintext literal. The provider must
+	// still store it as a secret: for a property nested inside a type the
+	// schema's `"secret": true` only reaches the .NET SDK, so for this Node.js
+	// program only Check keeps it out of the state file.
+	// Guards https://github.com/pulumi/pulumi-pulumiservice/issues/1037.
+	runPulumiTestAfterUp(t, test, func(t *testing.T, test *pulumitest.PulumiTest) {
+		inputs := stateInputs(t, test, "pulumiservice:index:DeploymentSettings")
+		assertStateSecret(t, inputs, "sourceContext", "git", "gitAuth", "sshAuth", "sshPrivateKey")
+		assertStateSecret(t, inputs, "sourceContext", "git", "gitAuth", "sshAuth", "password")
+	})
 }
 
 func TestTeamStackPermissionsExample(t *testing.T) {
--- a/examples/main_test.go
+++ b/examples/main_test.go
@@ -1,9 +1,13 @@
 package examples
 
 import (
+	"encoding/json"
 	"os"
 	"os/exec"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/pulumi/providertest/providers"
 	"github.com/pulumi/providertest/pulumitest"
@@ -11,6 +15,8 @@ import (
 	"github.com/pulumi/providertest/pulumitest/assertrefresh"
 	"github.com/pulumi/providertest/pulumitest/opttest"
 	"github.com/pulumi/pulumi/sdk/v3/go/auto"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 
 	psp "github.com/pulumi/pulumi-pulumiservice/provider/pkg/provider"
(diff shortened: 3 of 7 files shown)


lukebward (approved):
no blockers from me!

PR #1092, opened 2026-09-19
Stop Stack.forceDestroy from replacing the stack

### LINKED ISSUE #1093: Stack.forceDestroy causes stack resource replacement
## What happened

Any change to `pulumiservice.Stack.forceDestroy` is treated as a replacement of the stack. The flag is only consulted in `Delete` and is not stored by the Pulumi Cloud API, so flipping it should be a state-only update — but 1.x declares it `replaceOnChanges` (inherited from the 0.x resource's "every change replaces" `Diff`, carried over mechanically by the `infer` migration in #829). `AgentPool.forceDestroy`, the same flag on the sibling resource, is not tagged.

The replacement can't succeed either way: `CreateStack` returns 409 for a stack that already exists, so `pulumi up` aborts before anything downstream of the `Stack` (e.g. `DeploymentSettings`) is applied.

## Where it bites: upgrading from 0.x

You don't have to touch `forceDestroy` to hit this. Providers ≤ v0.41.0 wrote the flag to state only when `true`, while `infer` materializes the non-pointer `bool` as `false`. So every `Stack` created under 0.x with the flag unset diffs `forceDestroy: <absent> → false` on its first `up` with any 1.x provider:

```
+-pulumiservice:index:Stack: (replace)
  + forceDestroy: false
```

`pulumi refresh` does not clear it (`Read` copies the old input, which is absent), and `ignoreChanges: ["forceDestroy"]` does not either (the provider computes the diff). The only workaround is editing `"forceDestroy": false` into the `Stack` resource's inputs and outputs via `pulumi stack export` / `import`.

## Affected

- Any program that changes `forceDestroy` on an existing `Stack`, on any 1.x provider.
- Any `Stack` created by a 0.x provider, on its first `up` with a 1.x provider (v1.1.0–v1.3.0), with no program change at all.

kmosher:
## Summary

Fixes #1093.

- `Stack.forceDestroy` no longer replaces the stack when it changes; it is a state-only update.
- State written by a pre-v1.1.0 provider (which omitted a `false` flag) no longer diffs at all: `Diff` compares typed values, so the missing key decodes to `false` and matches.
- `Update` refuses identity changes, so a future input added without `replaceOnChanges` cannot become a silent state-only update.
- Regression test covers the pre-v1.1.0 upgrade path (no diff), a `true`/`false` flip (update), that identity properties still replace, and that `Update` calls no API.

## Why

`forceDestroy` is consulted only in `Delete`, and the Pulumi Cloud API does not store it (`Read` already treats it as a write-only hint). It was nonetheless a replacing property, for two compounding reasons:

- The original 0.x resource (#357) used a generic "any input change ⇒ replace, delete-before-replace" `Diff` written for the identity properties, and `forceDestroy` shipped inside it.
- The `infer` migration (#829, v1.1.0) carried that over as `provider:"replaceOnChanges"` on every field. `AgentPool`'s `forceDestroy` — same flag, same semantics — was migrated without it.

That would have been harmless except that 0.x providers wrote `forceDestroy` to state only when `true`, while `infer` materializes a non-pointer `bool` as `false`. So on the first `pulumi up` after upgrading a 0.x-created `Stack` to any 1.x provider, the diff is `forceDestroy: <absent> → false` on a replace-on-change property:

```
+-pulumiservice:index:Stack: (replace)
  + forceDestroy: false
```

The replacement then fails — `CreateStack` 409s on the existing stack — so the update aborts before any dependent resource (e.g. `DeploymentSettings`) is touched. Neither `pulumi refresh` (`Read` copies the absent input) nor `ignoreChanges: ["forceDestroy"]` (the provider computes the diff) clears it; the only workaround was patching `"forceDestroy": false` into state by hand. Every user with a `Stack` created before v1.1.0 hits this.

Dropping the tag alone is not enough: with no `Update` method, `infer` treats every diff as a replacement regardless of tags. `Update` records the new inputs and calls no API. And removing the replacement alone still left one `~ update forceDestroy: false` per pre-v1.1.0 stack — infer's default diff works on raw property maps, where a missing key differs from `false` — which would trip `preview --expect-no-changes` once per stack; the typed `Diff` closes that, following the precedent #752/#760 set for `TeamEnvironmentPermission`.

The remaining `replaceOnChanges` tags across the provider were audited as part of this; all others are identity properties, or API-immutable fields (access-token description/admin, permissions), or have a schema default that has always been sent by the SDK. The schedule `timestamp` fields are updatable via the API but replacing them is benign (a new schedule id); left as-is.

## Test plan

- `go test ./provider/pkg/resources/` — `TestStackForceDestroy` fails on the previous code with `add&replace` (upgrade path) and `update&replace` (flip); with only the tag removed and `Update` added it still fails the upgrade case with `{add false}`; passes with the typed `Diff`.
- Against a stack whose state was written by a 0.x provider, `pulumi preview` with this provider reports no changes; flipping `forceDestroy` in the program shows `~ 1 to update` and `up` converges state without calling the API.

https://claude.ai/code/session_01LSU1XaZwPXiy6uad9s6YND


provider/pkg/resources/stack.go
```diff
@@ -88,6 +94,55 @@ func (*Stack) Create(
 	}, nil
 }
 
+// Diff compares typed values rather than raw property maps so that state
+// written by a provider before v1.1.0, which omitted forceDestroy when false,
+// decodes to the same false the current inputs carry and produces no diff.
+func (*Stack) Diff(
```
iwahbe, line 100: Why is `Diff` necessary to implement? `Stack.ForceDestroy` isn't `provider:"replaceOnChanges"` and `Stack.Update` exists.

kmosher: Dropping the tag and adding `Update` is enough to stop the *replacement*, but not enough to stop the *diff*.

Comes up in the transition from 0.x to 1.x providers. The previous state was encoded as `<absent>/true`, while the new defaults are `false/true`, so the Diff prevents a spurious HasChanges and detected drift from the `<absent>->false` transition.

There's a test that enforces this: with the tag removed and `Update` added but no `Diff`, the upgrade-path subtest still fails with `{add false}`. Comparing typed values makes the missing key decode to `false` and the diff empty. Same shape as the normalization `TeamEnvironmentPermission` got in #752/#760.

<sub>🤖 posted from Keith's automation · Claude Opus 5 -- rewritten by hand</sub>

iwahbe (approved):
State migration is a better answer on https://github.com/pulumi/pulumi-pulumiservice/pull/1092/changes#r4063643927, but it won't cover migrating the old inputs, which is what the default diff compares on.

I'll file an issue.

